PR #3, opened 2023-06-07
fix: fix list relationship query filters

jgadling:
This updates the example code so that it doesn't surprise anyone with large downloads


chili-chiu (approved):
looks good, thanks!

PR #9, opened 2023-06-09
fix: Fix get_by_id

### LINKED ISSUE #5: Error in  Client.get_by_id
version == 1.0.0
```
from cryoet_data_portal import Client, Dataset
myclient=Client()
item = Dataset.get_by_id(myclient, 10000)
```
Gave error

```
  File "/Users/anchi.cheng/Library/Python/3.9/lib/python/site-packages/cryoet_data_portal/_gql_base.py", line 259, in get_by_id
    return client.find_one(cls, [cls.id == id])
  File "/Users/anchi.cheng/Library/Python/3.9/lib/python/site-packages/cryoet_data_portal/_client.py", line 58, in find_one
    for result in self.find(*args, **kwargs):
  File "/Users/anchi.cheng/Library/Python/3.9/lib/python/site-packages/cryoet_data_portal/_client.py", line 53, in find
    response = self.client.execute(self.build_query(cls, gql_type, query_filters))
  File "/Users/anchi.cheng/Library/Python/3.9/lib/python/site-packages/gql/client.py", line 403, in execute
    return self.execute_sync(
  File "/Users/anchi.cheng/Library/Python/3.9/lib/python/site-packages/gql/client.py", line 221, in execute_sync
    return session.execute(
  File "/Users/anchi.cheng/Library/Python/3.9/lib/python/site-packages/gql/client.py", line 860, in execute
    raise TransportQueryError(
gql.transport.exceptions.TransportQueryError: {'extensions': {'code': 'validation-failed', 'path': '$.selectionSet.datasets.args.where'}, 'message': "expected an object for type 'datasets_bool_exp', but found an enum value"}

```

Workaround:
```
    items = Dataset.find(client,[Dataset.id==10004])
    for item in items:
        print(item.title)
```

jgadling:
Closes #5

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
@@ -256,7 +256,9 @@ class Model:
             >>> run = Runs.get_by_id(client), 1
                 print(run.name)
         """
-        return client.find_one(cls, [cls.id == id])
+        results = cls.find(client, [cls.id == id])
+        for result in results:
+            return result
 
     @classmethod
     def setup(cls):


andy-sweet (approved):
Good enough for me as a quick fix.

To clarify, I think the underlying issue here is that `Client.find` and `Client.find_one` won't work if you pass the same `Iterable[GQLExpression]` as to `Model.find`.

Another way to think about that is that out of the following tests

```python
from cryoet_data_portal import Client, Dataset

def test_client_find_dataset() -> None:
    client = Client()
    datasets = tuple(client.find(Dataset))
    assert len(datasets) > 0

def test_client_find_dataset_with_id() -> None:
    client = Client()
    datasets = tuple(client.find(Dataset, [Dataset.id == 10000]))
    assert len(datasets) == 1

def test_dataset_find() -> None:
    client = Client()
    datasets = tuple(Dataset.find(client))
    assert len(datasets) > 0

def test_dataset_find_with_id() -> None:
    client = Client()
    datasets = tuple(Dataset.find(client, [Dataset.id == 10000]))
    assert len(datasets) == 1
```

only `test_client_find_dataset_with_id` fails. 

I was going to suggest a higher level fix (perhaps by using `to_gql` in `Client.find`), but then I realized that while `Client.find` is public, it has no docstring, thus isn't in the API reference, and also isn't mentioned in the docs at all.

So I guess it's not intended as part of the public API, so this fix is fine.

Issue #22, opened 2023-09-27
Browse all data (list view & pagination)


junxini:
**User story:** (2.1) I want to browse all data, so that I am aware of what is available

**Tasks** 
- [x] Show dataset in list view
- [x] pagination 

**Link to design:** https://www.figma.com/file/WEmbsjtlBUtRy7pzmuCCjj/CryoET-Data-Portal---Phase-2-Designs?mode=dev

junxini:
@codemonkey800 - here's the backend ticket for author ordering which would apply to the list view https://github.com/chanzuckerberg/cryoet-data-portal/issues/100

junxini:
@codemonkey800 - another backend story where you'll find the key photos #56

lauramarcos:
Im opening this one as Pagination cannot be tested in current staging site @codemonkey800

PR #88, opened 2023-10-12
browse all datasets page

codemonkey800:
| Issue | Depends on |
| ----- | ---------- |
| #22   | #87        |

## Description

Implements the browse all data page with the following functionality:

- Dataset / Runs tabs with counts
- Search by dataset title
- Empty filter sidebar
- Pagination
- Table view for datasets
  - Key photo
  - Link to single dataset
  - Author list + overflow
  - EMPIAR ID link
  - Organism name
  - Run count
  - Annotated object list UI + full list in tooltip on hover
  - Skeleton loader when navigating to another page

## Follow up work

- Error page
- No results page
- Use annotated objects from API
- Use dataset key photo

### Browse all datasets

image

### Datasets / Runs tabs with counts

image

### Table with pagination

image

### Table skeleton loader

https://github.com/chanzuckerberg/cryoet-data-portal/assets/2176050/2cbabb0f-a990-4b1c-ab8f-13e38213b362

### Annotation objects tooltip

image

### DIFF
--- a/frontend/packages/data-portal/app/components/BrowseData/BrowseDataSearch.tsx
+++ b/frontend/packages/data-portal/app/components/BrowseData/BrowseDataSearch.tsx
@@ -5,12 +5,18 @@ import { useState } from 'react'
 
 import { i18n } from 'app/i18n'
 
+/**
+ * The amount of time to wait after the user has typed in a query before
+ * re-fetching the list of datasets. The 500ms should be enough time for people
+ * of variable typing speeds to enter a complete query.
+ */
 const SEARCH_QUERY_DEBOUNCE_TIME_MS = 500
 
 export function BrowseDataSearch() {
   const [searchParams, setSearchParams] = useSearchParams()
   const [query, setQuery] = useState(searchParams.get('search') ?? '')
 
+  // If the user hasn't typed in a key for 500ms, then update the search params.
   useDebouncedEffect(
     () =>
       setSearchParams((prev) => {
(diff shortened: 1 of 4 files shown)


frontend/packages/data-portal/app/components/BrowseData/BrowseDataSearch.tsx
```diff
@@ -0,0 +1,39 @@
...
+          prev.set('search', query)
+        } else if (prev.has('search')) {
+          prev.delete('search')
+        }
+
+        return prev
+      }),
+    [query],
+    SEARCH_QUERY_DEBOUNCE_TIME_MS,
+  )
```
kne42, line 27: you went over why we need a debounce in the live code review, could you maybe add a comment somewhere here in case I forget later down the line?

frontend/packages/data-portal/app/components/BrowseData/DatasetKeyPhoto.tsx
```diff
@@ -0,0 +1,39 @@
...
+export function DatasetKeyPhoto({
+  datasetTitle,
+  src,
+}: {
+  datasetTitle: string
+  src?: string
+}) {
+  const { isLoadingDebounced } = useIsLoading()
+  const containerClassName =
+    'min-w-[134px] min-h-[100px] max-w-[134px] max-h-[100px]'
```
kne42, line 16: probably I just dont know tailwind well enough but why are we not just setting the width directly instead of including a min and max that are the same value?

frontend/packages/data-portal/app/routes/browse-data.runs.tsx
```diff
@@ -0,0 +1,5 @@
+import { Demo } from 'app/components/Demo'
+
+export default function BrowseRunsPage() {
+  return <Demo>Browse runs page</Demo>
+}
```
kne42, line 5: since this is a demo, maybe add a TODO to replace it with the real thing later? (idk just nitpicking)

kne42 (approved):
overall LGTM! left a few non-blocking comments

PR #92, opened 2023-10-13
refactor live reload UI

codemonkey800:
## Description

Refactors live reload to include more states and fix a few bugs with hot module reloading (HMR).

### Custom Dev Server

A custom dev server was used to fix HMR due to a bug with Remix v2. A current issue exists regarding the issue with HMR here: https://github.com/remix-run/remix/issues/7466.

A current workaround to the issue is to use a custom dev server as described [in this comment](https://github.com/remix-run/remix/issues/7466#issuecomment-1751996791).

The sever implementation is based on a [community example for using ES Modules + TypeScript](https://github.com/xHomu/remix-v2-server/tree/esm-server.ts).

### Added file

image

### Deleted file

image

### Modified file

Unlike when a file is added or deleted, the overlay for modified files only disappear when an HMR event is received from the server. That's why when the overlay disappears, the UI is immediately updated. If we did it when the bundle was finished building, there would be a small delay between the UI being rebuilt and being rerendered.

https://github.com/chanzuckerberg/cryoet-data-portal/assets/2176050/c2d81d16-ef41-4c36-a82a-f69aec471d71

### Multiple files

Editing multiple files during an in progress live reload will prolong the reload, so it'd be nice to know which files were updated during the reload on the overlay screen.

https://github.com/chanzuckerberg/cryoet-data-portal/assets/2176050/b85c410e-7c55-449b-92e6-eb3809a23187

### Live reload taking long

Sometimes live reload will appear stuck due to an error that is only visible within the terminal. Because of this, an additional warning is shown when live reload takes a long time to complete.

image

### Websocket closed

When the dev server is closed, the frontend will lose it's connection to the dev server's websocket connection. We show an overlay when this happens to make it clear that the dev server isn't running.

image

### DIFF
--- a/frontend/packages/data-portal/package.json
+++ b/frontend/packages/data-portal/package.json
@@ -14,7 +14,7 @@
     "lint": "eslint --report-unused-disable-directives --max-warnings 0 .",
     "lint:fix": "pnpm lint --fix",
     "preinstall": "npx only-allow pnpm",
-    "start": "NODE_ENV=production node --lodaer ts-node/esm server.ts",
+    "start": "NODE_ENV=production node --loader ts-node/esm server.ts",
     "test": "jest"
   },
   "dependencies": {


kne42 (approved):
LGTM. might be a dumb question but is there not a package that already has live reload functionality?

PR #96, opened 2023-10-17
linting + CI

kne42:
adds linting (including pre-commit) and workflows for running linting and tests:
- pre-commit
- frontend tests
- frontend linting
- frontend type-check
- backend linting
- backend code-style


pyproject.toml
```diff
@@ -0,0 +1,34 @@
+# common tooling configs
+[tool.black]
```
manasaV3, line 2: Curious about why this was replicated here in addition to `client/python/cryoet_data_portal`? 😅

py-dev-requirements.txt
```diff
@@ -0,0 +1,2 @@
+black
```
manasaV3, line 1: Could this file also be a part of the `client/python/cryoet_data_portal`?

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
```diff
@@ -18,35 +21,42 @@ def parse_s3_url(url: str) -> (str, str):
...
+) -> str:
     if not dest_path:
         dest_path = os.getcwd()
     dest_path = os.path.abspath(dest_path)
 
-    if not recursive_from_prefix:
-        # If the dest path is a directory, append the url's file name to it.
-        if os.path.isdir(dest_path):
-            return os.path.join(dest_path, os.path.basename(url))
     if not os.path.isdir(dest_path):
```
manasaV3, line 54: Looks like this might be more than linting changes. This probably needs the latest commits to be pulled in?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    files: frontend
+- repo: local
+  hooks:
+  - id: pnpm-lint
+    name: pnpm lint
+    entry: pnpm -r --parallel lint:fix
+    language: system
+    types_or: [javascript, jsx, ts, tsx]
+    files: frontend
+    pass_filenames: false
```
manasaV3, line 27: Nit: needs a new line at the end of the file.

manasaV3 (changes requested):

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
```diff
@@ -18,35 +21,42 @@ def parse_s3_url(url: str) -> (str, str):
...
+    url: str,
+    dest_path: Optional[str] = None,
+    with_progress: bool = True,
+):
     dest_path = get_destination_path(url, dest_path)
     fetch_request = requests.get(url, stream=True)
     total_size = int(fetch_request.headers["content-length"])
     block_size = 1024 * 512
-    print(f"Downloading {url} to {dest_path}")
+    logger.info("Downloading %s to %s", url, dest_path)
```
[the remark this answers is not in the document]
kne42: formatted strings in log statements upset the linters because technically python still has to convert the objects to strings even if the log is not run

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  hooks:
+  - id: ruff
+    files: backend|client/python
+    args:
+    - --fix
+- repo: https://github.com/pre-commit/mirrors-prettier
+  rev: v3.0.3
+  hooks:
+  - id: prettier
+    files: frontend
```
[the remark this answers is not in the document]
kne42: this does use `npm install` I would assume. we can circumvent this by deleting it and letting the `pnpm lint` command run it instead

client/python/cryoet_data_portal/pyproject.toml
```diff
@@ -46,10 +46,6 @@ exclude = ["tests*"]  # exclude packages matching these glob patterns (empty by
 [tool.setuptools_scm]
 root = "../../.."
 
-[tool.black]
-line-length = 120
```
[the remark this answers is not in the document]
kne42: i can revert this if you desire

pyproject.toml
kne42: `pre-commit` searches for the tooling settings under the root directory; also this means we can have one configuration for both the backend and client folders (which can be overridden by their separate `pyproject.toml`s

.github/workflows/formatting.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  pre-commit-checks:
+    name: pre-commit checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
```
manasaV3, line 22: Nit: Do we need this?

client/python/cryoet_data_portal/pyproject.toml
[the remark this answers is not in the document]
manasaV3: Not a blocker: Yes, please, if it isn't too much trouble.

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
kne42: yes, these aren't automatic changes, these were ones I made myself after receiving the following linting error:
```python
client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py:47:5: SIM102 Use a single `if` statement instead of nested `if` statements
```

I basically just moved around the logic but the functionality _should_ stay the same. Does it look to you like the functionality will be the same or is there another way you think it should be written? Of course this would be easy to verify if we had unit tests...

.github/workflows/formatting.yml
kne42: technically not, no

manasaV3: Not a blocker: We could remove it.

manasaV3 (approved):

kne42:
the workflow is failing for reasons i cannot reproduce locally... id rather just merge it in now and maybe we can fix it later since it's mainly meant as a warning

PR #101, opened 2023-10-20
single dataset route

codemonkey800:
## Description

Implements the `/datasets/$id` route with the header and back navigation implemented. The back navigation works by passing the previous URL as a query parameter. This simplifies the back to search results functionality because it allows us to keep the previous URL data stateless.

## Demo

https://github.com/chanzuckerberg/cryoet-data-portal/assets/2176050/4b6eab3b-2a62-4f20-8e3c-4bee2f0c8778

### DIFF
--- a/frontend/packages/data-portal/app/components/BrowseData/DatasetTable.tsx
+++ b/frontend/packages/data-portal/app/components/BrowseData/DatasetTable.tsx
@@ -164,7 +164,7 @@ export function DatasetTable() {
               .map((val) => `Object ${val}`)
               .map((obj) => <li key={obj}>{obj}</li>)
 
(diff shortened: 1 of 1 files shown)


frontend/packages/data-portal/app/routes/datasets.$id.tsx
```diff
@@ -0,0 +1,101 @@
...
+      last_modified_date
+      release_date
+    }
+  }
+`)
+
+export async function loader({ params }: LoaderFunctionArgs) {
+  const id = params.id ? +params.id : NaN
+
+  if (Number.isNaN(+id)) {
```
kne42, line 29: what does `+id` mean?

kne42 (approved):

Issue #31, opened 2023-09-28
key photo at dataset level


junxini:

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/56

lauramarcos:
key photo in All data page is not displaying dataset page https://github.com/chanzuckerberg/cryoet-data-portal/issues/137

PR #117, opened 2023-10-31
add single dataset key photo

kne42:
Implemented as in #31. Adds the dataset level key photo. Note: actual key photo not available from database yet. Contains CSS code which does not run on IE (which when synced with @codemonkey800 is not a browser we support anyways).

https://github.com/chanzuckerberg/cryoet-data-portal/assets/29165011/fc18e9c5-4e2a-4732-8bb7-11217ea6e33e


kne42:
> hmm you might need to rebase from main and remove my commits:
> 
> image

yeah i just had to force-push my other branch; updated this one as well

PR #119, opened 2023-11-04
run page header

codemonkey800:
## Description

Implements the single run page header. This refactors the code to reuse the same page header code from the single dataset page. This also sets up the initial query for fetching a run by ID.

## Demo

image

## Next Steps

- Add data for frames

### DIFF
--- a/frontend/packages/data-portal/app/components/Dataset/DatasetHeader.tsx
+++ b/frontend/packages/data-portal/app/components/Dataset/DatasetHeader.tsx
@@ -23,7 +23,13 @@ export function DatasetHeader() {
         </Button>
       }
       lastModifiedDate={dataset.last_modified_date ?? dataset.deposition_date}
-      metadata={[{ key: i18n.portalIdBlank, value: String(dataset.id) }]}
+      metadata={[
+        {
+          key: i18n.portalIdBlank,
+          value: String(dataset.id),
+          uppercase: true,
+        },
+      ]}
       onMoreInfoClick={drawer.toggle}
       releaseDate={dataset.release_date}
       title={dataset.title}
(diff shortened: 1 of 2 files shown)


frontend/packages/data-portal/app/i18n.ts
```diff
@@ -58,13 +63,14 @@ export const i18n = {
   plusMore: (count: number) => `+${count} More`,
   poor: 'Poor',
   portalId: (id: number | string) => `Portal ID: ${id}`,
-  portalIdBlank: 'Portal ID:',
+  portalIdBlank: 'PORTAL ID',
```
kne42, line 66: I'd like this to remain title case but use `text-transform: uppercase` in the css

frontend/packages/data-portal/app/i18n.ts
```diff
@@ -78,10 +84,17 @@ export const i18n = {
   submitFeedback: 'Submit Feedback',
   terms: 'Terms',
   termsOfUse: 'Terms of Use',
+  tiltQuality: 'Tilt Quality',
+  tiltRange: 'Tilt Range',
+  tiltRangeValue: (min: number, max: number) => `${min}° to ${max}°`,
```
kne42, line 89: I have another version of this but more generalized in my PR, what are your thoughts on it?

frontend/packages/data-portal/app/state/drawer.ts
```diff
@@ -2,10 +2,10 @@ import { useUnmountEffect } from '@react-hookz/web'
 import { atom, useAtom, useSetAtom } from 'jotai'
 import { useMemo } from 'react'
 
-const datasetDrawerOpenAtom = atom(false)
+const drawerOpenAtom = atom(false)
 
-export function useDatasetDrawer() {
-  const [open, setOpen] = useAtom(datasetDrawerOpenAtom)
+export function useDrawer() {
```
kne42, line 7: glad we had the same idea yippie! (maybe I am doing things right!)

kne42 (approved):

PR #136, opened 2023-11-09
Tutorial on using AWS CLI and Github Issue

dgmccart:


anchi2c (changes requested):
This is too long for user to read without wandering off to aws pages and end up on the wrong track.  Please keep it short with the important info up front:  install but no need to go through configuration for role.

anchi2c (commented):
I still prefer the first paragraph states that cryo-et data portal s3 bucket is public and requires no credential sign-in to download.  You can then create a link from there to the part of with more detail description of using --no-sign-in-requests

 Please let me know if there are reasons that such statement up-front is not a good idea.

 Also, I think the part about people may want to configure credential should be put as a separate paragraph since it is purely optional and not part of our recommended path.

Thanks.

dgmccart:
Hi Anchi, thank you for clarifying how to simplify. I added a paragraph at the start of the guide:

"
The Data Portal's S3 bucket is public, so it can be used without credential sign in by specifying --no-sign-request in your commands. In only a few minutes, you can get started downloading data following the steps in the [Quickstart Guide](https://github.com/chanzuckerberg/cryoet-data-portal/blob/dgmccart-patch-1/docs/cryoet_data_portal_docsite_aws_download.md#quickstart), and for more details, refer to the [Installation](https://github.com/chanzuckerberg/cryoet-data-portal/blob/dgmccart-patch-1/docs/cryoet_data_portal_docsite_aws_download.md#installation), [Download Data](https://github.com/chanzuckerberg/cryoet-data-portal/blob/dgmccart-patch-1/docs/cryoet_data_portal_docsite_aws_download.md#download-data), and [Optimize Download Speed](https://github.com/chanzuckerberg/cryoet-data-portal/blob/dgmccart-patch-1/docs/cryoet_data_portal_docsite_aws_download.md#optimize-download-speed) in-depth explanations.
"

anchi2c (approved):

Issue #111, opened 2023-10-26
Single Run Page - Annotation Metadata 


junxini:
User story: (4.2) I want to see all the annotation metadata from single run page, so that I can access detailed information

Tasks

- [ ] full list of annotation metadata accessible but initially tucked away (likely via panel) - open via "more info" within each row within the table 
- [ ] when panel is open, the rest of the page (outside of the panel) is greyed out and not accessible - user to click on "X" or somewhere outside of the panel to close
- [ ]  the annotation row that metadata is opened from is highlighted in gray

Link to design: https://www.figma.com/file/WEmbsjtlBUtRy7pzmuCCjj/CryoET-Data-Portal---Phase-2-Designs?type=design&node-id=1310%3A11321&mode=dev

Janeece:
Screenshot from Design:


![Image](https://github.com/chanzuckerberg/cryoet-data-portal/assets/63013360/714389b6-42c4-4c55-a57b-fb0a029985ae)

lauramarcos:
Following bug was found during testing:

Page not greyed out and accesible when right panel is opened [#245](https://github.com/chanzuckerberg/cryoet-data-portal/issues/245)

PR #155, opened 2023-11-15
single run annotation metadata

codemonkey800:
#111

- Refactors code to use [remix-typedjson](https://github.com/kiliman/remix-typedjson) for better typing when using loader data
- Add state for active annotation data
- Implements new components for the annotation drawer

## Demo

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/Matrix4x4.tsx
+++ b/frontend/packages/data-portal/app/components/Run/Matrix4x4.tsx
@@ -1,11 +1,10 @@
-import { TableDataValue } from 'app/types/table'
 import { cns } from 'app/utils/cns'
 
 export function Matrix4x4({
   matrix,
   className,
 }: {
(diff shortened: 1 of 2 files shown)


frontend/packages/data-portal/app/components/Run/Matrix4x4.tsx
```diff
@@ -1,10 +1,11 @@
+import { TableDataValue } from 'app/types/table'
 import { cns } from 'app/utils/cns'
 
 export function Matrix4x4({
   matrix,
   className,
 }: {
-  matrix: string
+  matrix: TableDataValue
```
kne42, line 8: why this change? it was meant as a string only because the way i encoded the matrix was a flat string delimited by spaces e.g. `1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1`

kne42 (approved):

PR #163, opened 2023-11-22
update frontend to work with staging data

codemonkey800:
Updates the frontend to work with the updated annotations schema where the shape type and s3 paths are stored in the `files` property.

For the annotations table, this makes it so that the files within each annotation are rendered as separate rows within the table. This doesn't change how we handle pagination for the annotation table, pagination is still handled with `limit` and `offset` at the `annotations` level, so this means there might be bugs in how we paginate on the table.

For example, given the pagination limit of 20 items, if the total number of files for a list of 20 annotations is >20, then the table will show the entire list of files which would break the pagination limit. With this in mind, we'll need to figure out a way to do some kind of pagination that operates on the flat list of files


manasaV3 (approved):
🔥

manasaV3:
I wonder if we can leverage schema resolvers and cursors to work around the issue with pagination.

PR #164, opened 2023-11-22
Update cryoet_data_portal_docsite_landing.md

dgmccart:
Correcting a statement. Not all data in the portal will be provided with point annotations.


uermel (approved):
This seems fine to me.

Issue #157, opened 2023-11-18
UXR study - minor updates for Dec beta release


junxini:
3 small suggestions we're hoping to tweak based on UXR interviews. The interviews are still in progress but there has been enough feedback on the following items that we feel confident moving forward with them.

In prioritized order:
- [ ] add 5 tool tips, in 7 locations. tool tip design and copy [here](https://www.figma.com/file/WEmbsjtlBUtRy7pzmuCCjj/CryoET-Data-Portal---Phase-2-Designs?node-id=2674%3A14097&mode=dev)
  - [ ] For "Tilt Series Quality" tooltip, there are some small updates (copy, column width) to the labels the tooltips attach to on Single Dataset and Single Run pages.
  - [ ] learn more links are dependent on FAQ copy -- if we do not have FAQ section written, do not include link. (will update if I learn ahead of time to not include)
  - [ ] [SDS tooltip component docs](https://sds.czi.design/009eaf17b/p/74af45-tooltips/t/84dfe9)
- [ ] add a banner to the feedback survey on all pages. feedback survey design [here](https://www.figma.com/file/WEmbsjtlBUtRy7pzmuCCjj/CryoET-Data-Portal---Phase-2-Designs?node-id=2754%3A25272&mode=dev)
  - [ ] if possible, allow users to dismiss for 2 weeks at a time. if eng bandwidth doesn't allow for this, having the button be not dismissible is fine, just limit it to the Browse all data page
- [ ] add a redundant "more info" button that opens up the metadata panel on the single dataset and single run page. button design[ here](https://www.figma.com/file/WEmbsjtlBUtRy7pzmuCCjj/CryoET-Data-Portal---Phase-2-Designs?node-id=2674%3A14100&mode=dev)
  - [ ] Uses minimal button to keep number of outlined buttons limited. [SDS minimal button component doc](https://sds.czi.design/009eaf17b/p/47778c-buttons/t/616c45)
- [ ]  remove the number of annotated objects on the single dataset page "runs" table. the object name would stay, just the numbers would be omitted
  - [ ] Column will look the same as it does on All Datasets page

junxini:
@kandarpksk - could you help update this ticket once the tooltips are ready? 
@Janeece - do you have advice on where that button is added and what it looks like?

junxini:
> design and copy is still pending

@kandarpksk to update ticket with newest link and design

kandarpksk:
[...]

PR #229, opened 2023-12-07
uxr minor updates

codemonkey800:
#157

- Adds new tooltips + Tooltip component and utils
- Adds survey banner
- Add redundant more info button

### Tooltips

image

image

image

image

image

image

image

### Survey Banner

image

### Redundant more info button

image

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/RunHeader.tsx
+++ b/frontend/packages/data-portal/app/components/Run/RunHeader.tsx
@@ -91,7 +91,7 @@ export function RunHeader() {
       title={run.name}
       renderHeader={({ moreInfo }) => (
         <div className="flex gap-sds-xxl p-sds-xl border-t-[3px] border-sds-gray-200">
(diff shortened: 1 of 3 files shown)


frontend/packages/data-portal/app/components/Run/RunHeader.tsx
```diff
@@ -88,78 +89,87 @@ export function RunHeader() {
...
+              <Link to={keyPhotoURL}>
+                <KeyPhoto title={run.name} src={keyPhotoURL} />
+              </Link>
+            ) : (
+              <KeyPhoto title={run.name} />
+            )}
+          </div>
+
+          <div className="flex flex-col gap-sds-xl">
+            <div className="flex flex-1 gap-sds-xxl flex-col lg:flex-row">
```
kne42, line 105: it seems github didnt actually save it but my comment was to move `flex-1` up a div since in your screenshot, the key photo height for the single runs page doesnt match that of the table + button as in the design

PR #252, opened 2023-12-12
hook up affine transformation matrix field

kne42:


frontend/packages/data-portal/codegen.ts
```diff
@@ -20,6 +20,7 @@ const config: CodegenConfig = {
         scalars: {
           date: 'string',
           numeric: 'number',
+          _numeric: 'number[][]',
```
richaagarwal, line 23: Why is this named `_numeric`?

Issue #59, opened 2023-10-03
Download a single dataset


junxini:
User story: (5.1) I want to download a single dataset, so that I can use it to develop/train ML models

**Tasks:**

P0:
- [ ] download button on single dataset page to open download dataset dialog
- [ ] dialog contains the following elements: title, dataset name, and 2 download tabs for users to chose from 
- [ ] tab 1: download via AWS S3
- _- step 1 asks user to select their save destination (this is optional for user) - P1_ 
- step 2 is a AWS s3 snippet that user can copy
- provide link to AWS instructions 
  - note: instructions are WIP by Dannielle
- [ ] tab 2: download via API
- info box to link user to API documentation
  - link is: https://chanzuckerberg.github.io/cryoet-data-portal/python-api.html
  
P1: 
- [ ] surface dataset ID for user to copy 
- [ ] asks user to select their save destination (this is optional for user)

Link to design: https://www.figma.com/file/WEmbsjtlBUtRy7pzmuCCjj/CryoET-Data-Portal---Phase-2-Designs?type=design&node-id=2315-169422&mode=design&t=8myNEdr3dZigyAP2-0

manasaV3:
Field to referece for source uri: `datasets.s3_prefix`

jgadling:
The download commands will be 

```
aws s3 --no-sign-request sync s3://cryoet-data-portal-public/10000/TS_030/TiltSeries/TS_030.zarr TS_030.zarr
```

Janeece:
Screenshots from design:

AWS tab:
![Image](https://github.com/chanzuckerberg/cryoet-data-portal/assets/63013360/9a223280-e911-4b49-9050-339931f5e499)

API tab:
![Image](https://github.com/chanzuckerberg/cryoet-data-portal/assets/63013360/17f1ebf3-8b42-43b5-8794-931e5f72ddc6)

lauramarcos:
Following bugs were found and raised for this ticket:

Dataset # instead of Dataset Name in download single dataset modal [#231](https://github.com/chanzuckerberg/cryoet-data-portal/issues/231)

Download Dataset button is missing the "..." at the end of label [#232](https://github.com/chanzuckerberg/cryoet-data-portal/issues/232)

Able to close "Download Dataset" modal by clicking outside [#233](https://github.com/chanzuckerberg/cryoet-data-portal/issues/233)

[...]

PR #262, opened 2023-12-13
Update AWS download command

richaagarwal:
Adds `--no-sign-request` flag so that AWS does not attempt to load credentials, as intended in https://github.com/chanzuckerberg/cryoet-data-portal/issues/59#issuecomment-1796480671.  Also adds a destination folder. 

Otherwise, the command fails with `fatal error: Unable to locate credentials`

**Note: This will still not work against staging due to permissions, but will work on the prod site.**

### DIFF
--- a/frontend/packages/data-portal/app/components/Download/AWSDownloadTab.tsx
+++ b/frontend/packages/data-portal/app/components/Download/AWSDownloadTab.tsx
@@ -24,7 +24,7 @@ export function AWSDownloadTab() {
     .with(['runs', DownloadConfig.AllAnnotations], () => s3AnnotationsPrefix)
     .otherwise(() => s3TomogramPrefix)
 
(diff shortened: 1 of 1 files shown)


jgadling (approved):

jgadling (changes requested):
Actually it would be *best* if instead of the `.` we used the last part of the ${s3Path}

richaagarwal:
@jgadling Do you know if the s3 path will always have a trailing `/` as it currently seems to on staging?  I can update it to get the 2nd to last element in that case (a little hacky but it's fine) but wanted to confirm first.

So, instead of `aws s3 --no-sign-request sync s3://cryoet-data-portal-staging/10003/ .` we would see `aws s3 --no-sign-request sync s3://cryoet-data-portal-staging/10003/ 10003`. That's the intention, right?

PR #281, opened 2023-12-14
fix: change type name in diagram

jgadling:


andy-sweet (approved):
Sorry for missing this! Thanks!

PR #282, opened 2023-12-14
Update Python API reference docs based on latest GraphQL schema

andy-sweet:
The main contribution of this PR is to update the Python API reference docs with the new classes added in the recent schema change (`TomogramAuthor` and `AnnotationFile`).

In addition, there are few other documentation related changes.

1\. Updated quick start to reflect other API changes (e.g. removing `binning` keyword argument from `download_mrcfile`).

2\. Added the upgrade option to the napari plugin, so that its dependencies (like the Python client in this repo) are upgraded when there is an existing installation.

3\. Fixed the examples in the base model so that they work with the current schema.

4\. Fixed some typos and small mistakes in the model docstrings.

5\. Enforced a version of Sphinx pre v7 to prevent an error when building the HTML docs.

The error is `Reason: UndefinedError("'style' is undefined")` and is likely related to the Sphinx RTD theme not supporting Sphinx v7. The [corresponding issue](https://github.com/readthedocs/readthedocs.org/issues/10279) should be fixed for v2 of the theme, but I did not find that to be the case.


jgadling (approved):
Thanks!!

PR #286, opened 2023-12-16
Remove 'Tomogram' from Available Files filter

richaagarwal:
Removing `Tomogram` from the Available Files filters (as at least a stopgap) due to the query timing out.


kandarpksk:
LGTM!

Issue #215, opened 2023-12-06
[staging] General Error flashing in frontend


lauramarcos:
**Description**
There are a few of these that happen as a result of some of the data not being available when the UI is ready to render, but we're not properly handling the data not being available but when the data becomes available then the application re-renders which is why there's an error flash

**Steps to Reproduce**
1. Click on any Dataset listed
2. Click on "Back to Results" link
3. Pay attention to image

**Actual Results**
Application error image flashed when coming back to Browse Data page

**Expected Results**
This error flashing should not happened

**Evidence**

https://github.com/chanzuckerberg/cryoet-data-portal/assets/63799148/54f70cc4-ab94-4803-b45f-e0733fa22e9c

lauramarcos:
@codemonkey800 I experienced a couple of these flashing bugs today but I am not certain if I will find them again.. I will leave this open for now and keep checking... I cannot say this is fixed yet tho!

lauramarcos:
@codemonkey800 I am still seeing these flashing errors, this evidence is from today


https://github.com/chanzuckerberg/cryoet-data-portal/assets/63799148/e07aa613-a78a-4647-b395-a5f583c87ccf

lauramarcos:
Happening when clicking on "Learn More" in Precision tooltip


https://github.com/chanzuckerberg/cryoet-data-portal/assets/63799148/68fb8a42-92b9-4486-ad9d-18482eb731b9

lauramarcos:
In production:

https://github.com/chanzuckerberg/cryoet-data-portal/assets/63799148/c548ea6b-8cca-44ef-9cb9-c9cee4cbdfe4

jgadling:
This still happens fairly often for me in **firefox** and it doesn't just flash, but I get the error and I have to click "back" to proceed:
![Screenshot 2023-12-18 at 4 26 13 PM](https://github.com/chanzuckerberg/cryoet-data-portal/assets/475208/4fe6f130-707a-4f61-8ea9-ee463bf88d48)

junxini:
From the FAQ page, i clicked on "Browse Data" in the top navigation bar today and got this error that persisted:

https://github.com/chanzuckerberg/cryoet-data-portal/assets/98626158/6cec3397-91a3-4df7-82f6-ae55cdbf6b98

lauramarcos:
this no longer reproducible, hopefully does not show up again.. thank you all!!

PR #328, opened 2024-01-10
browse datasets error boundaries

codemonkey800:
Adds error boundaries to various components on the browse datasets page. This is to help identify the issue for the `insertNode` error that happens intermittently (see #215).

This should also improve the UX around the error by preventing the entire application from crashing, and allowing the user to refresh the error via a button.

This also sets up server side logging for errors that happen within the error boundary so that any errors that aren't captured as a result of this are logged somewhere in CloudWatch so that we can inspect it later.

**This does not fix the error, this is to purely help with identifying the issue + improving UX around the error**

## Demo

https://github.com/chanzuckerberg/cryoet-data-portal/assets/2176050/eed7d29e-4894-4c67-9804-8bfefe110088

image

### DIFF
--- a/frontend/packages/data-portal/app/components/ErrorBoundary.tsx
+++ b/frontend/packages/data-portal/app/components/ErrorBoundary.tsx
@@ -22,7 +22,6 @@ function FallbackRender({ error, resetErrorBoundary }: FallbackProps) {
 
   useEffect(() => {
     if (logId) {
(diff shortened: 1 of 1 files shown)


manasaV3 (approved):
lgtm! 🚢

kne42 (approved):
nice, this should improve UX a lot. is there any way to see the logs when e.g. prod errors?

kne42:
awesome

PR #334, opened 2024-01-17
add more error boundaries to browse datasets page

codemonkey800:
#215

- Adds `getErrorMessage()` util function
- Updates `sendLogs()` to accept `...logs: LogEntry[]` instead of `logs: LogEntry[]`
- Moves table error boundaries to `Table` component
- Add `sendLogs()` for column creation


kne42 (approved):
LGTM!

Issue #34, opened 2023-09-28
all runs associated with dataset (list view)


junxini:
**User story:** (3.5) I want to see all runs associated with this dataset + filter/sort within the dataset so that I can easily navigate large amounts within dataset

**Tasks** 
- [ ] filter within a dataset page to narrow down on runs [Full list here](https://docs.google.com/spreadsheets/d/16mjJEitfU1rPD4wwyjD4e1j230gXvAic9iTvJFDB9DA/edit?pli=1#gid=1927013063)
- [ ] sort by name, tilt series quality, annotated object count (see designs for final list)

**Link to design:**

Janeece:
Screenshot from design ( for general reference):

![Image](https://github.com/chanzuckerberg/cryoet-data-portal/assets/63013360/8f222b9a-0873-4aae-aa33-2d612d412269)

PR #407, opened 2024-01-24
single dataset filter

codemonkey800:
#34
#456

- Adds a filter panel to the single dataset page
  - Quality score filter
  - Tilt range filter
- Fixes bug with tilt range not using `AND` correctly when min and max is defined

## Demos

https://dev-filter-test.cryoet.dev.si.czi.technology/browse-data/datasets

image

image

image

image

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Filters/TiltRangeFilter.tsx
+++ b/frontend/packages/data-portal/app/components/Filters/TiltRangeFilter.tsx
@@ -76,7 +76,7 @@ export function TiltRangeFilter() {
         .with([P.not(''), ''], () => [
           {
             value: t('valueToValue', {
-              value1: t('unitDegree', { value: tiltMin }),
+              value1: t('unitDegree', { value: tiltMinParam }),
               value2: t('unitDegree', { value: DEFAULT_TILT_RANGE_MAX }),
             }),
           },
@@ -85,15 +85,15 @@ export function TiltRangeFilter() {
           {
             value: t('valueToValue', {
               value1: t('unitDegree', { value: DEFAULT_TILT_RANGE_MIN }),
-              value2: t('unitDegree', { value: tiltMax }),
+              value2: t('unitDegree', { value: tiltMaxParam }),
             }),
           },
         ])
         .with([P.not(''), P.not('')], () => [
           {
             value: t('valueToValue', {
-              value1: t('unitDegree', { value: tiltMin }),
-              value2: t('unitDegree', { value: tiltMax }),
+              value1: t('unitDegree', { value: tiltMinParam }),
+              value2: t('unitDegree', { value: tiltMaxParam }),
             }),
           },
         ])
@@ -108,17 +108,13 @@ export function TiltRangeFilter() {
       }}
       onCancel={() => {
         setTiltMin(tiltMinParam)
-        setTiltMin(tiltMaxParam)
+        setTiltMax(tiltMaxParam)
       }}
       onRemoveFilter={() => {
         updateValues({
           [QueryParams.TiltRangeMin]: null,
           [QueryParams.TiltRangeMax]: null,
         })
-      }}
-      onOpen={() => {
-        setTiltMin('')
-        setTiltMax('')
       }}
     >
       <div className="flex flex-col gap-sds-xs max-w-[320px] mt-sds-xs">


frontend/packages/data-portal/app/utils/filter.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function getTiltRangeFilter(
+  initialTiltMin: string,
+  initialTiltMax: string,
+) {
+  let tiltMin = getTiltValue(initialTiltMin)
+  let tiltMax = getTiltValue(initialTiltMax)
+
+  if (isNumber(tiltMin) && !isNumber(tiltMax)) {
+    tiltMax = DEFAULT_TILT_MAX
```
manasaV3, line 22: Curious on how we arrived at the `DEFAULT_TILT_MAX` and `DEFAULT_TILT_MIN` values?

frontend/packages/data-portal/app/components/RunFilter/RunFilter.tsx
```diff
@@ -0,0 +1,19 @@
+import { TiltRangeFilter } from 'app/components/Filters'
+import { useI18n } from 'app/hooks/useI18n'
+
+import { QualityScoreFilter } from './QualityScoreFilter'
+
+export function RunFilter() {
+  const { t } = useI18n()
+
+  return (
+    <div className="px-sds-l flex flex-col gap-1">
```
kne42, line 10: use `gap-sds-xxxs` here instead

kne42 (approved):
one nit otherwise LGTM!

frontend/packages/data-portal/app/utils/filter.ts
manasaV3: @Janeece, sorry if this was addressed elsewhere; the minimum angle values can be negative, so we are being restrictive with our default value of 0. 

What do you think about either setting min and max defaults dynamically from the DB or not adding them to the filter when not explicitly specified?

Janeece: This should be consistent with how we implemented it on the all datasets filter. How is that one working?

HOWEVER, I think its good to question this filter — its design is proving to be confusing and i think i finally understand why it doesn't line up. 

Tilt range is the angle between the min and max tilt angles (-60min tilt angle to 60 max title angle = 180° tilt range). 

The filter can either (1) let users pick a single tilt range, like 60° tilt range, or (2) let users pick a range of tilt ranges, like 20° to 180° tilt range. 

1. Would result in a narrow filter, only one tilt range some value between 0° and 180° (ex filter by 180° tilt range). The min/max filter inputs in this case would be the min and max tilt angles (could be negative, like -60° to 60°)
2. Would let users filter by a range of tilt ranges, so any range between the min and max tilt range would be ok. The min/max filter inputs would be the minimum tilt range to the maximum tilt range (must be positive, like 0 to 180°). 

My understanding is that this filter should be accomplishing (2). But the copy (and maybe the form structure) are actually describing (1)  

I need to re-confirm with which it is with Anchi/Utz to know how to correct the confusion.

Janeece: Additionally, on the single run page we present tilt range as the min to max angle, not the range total

Janeece:
Design feedback:

### General
- seems like the spacing and type size issues kandarp called out are resolved

### Tilt Range filter
- [ ] Critical - When i try to remove an applied tilt range a 0° to 0° filter is applied instead of removeing the filter (behaviorally, this resets range to 0° to 180°) 
Screenshot 2024-02-01 at 11 48 00 AM
- [ ] I need to look into the [confusing parts of the design](https://github.com/chanzuckerberg/cryoet-data-portal/pull/407#discussion_r1475058207) with this filter, but it is behaving how i expect. 

### Tilt Series Quality Score
- [ ]Low severity - The menu item check is not correctly aligned 
Screenshot 2024-02-01 at 11 55 34 AM
- [ ] Low/medium severity - selection tag is displaying next to the filter instead of below 
Screenshot 2024-02-01 at 11 55 45 AM

frontend/packages/data-portal/app/utils/filter.ts
```diff
@@ -0,0 +1,42 @@
+import { isNumber } from 'lodash-es'
+
+import { Runs_Bool_Exp } from 'app/__generated__/graphql'
+import { DEFAULT_TILT_MAX, DEFAULT_TILT_MIN } from 'app/constants/tiltSeries'
```
manasaV3, line 4: Nit: Could these constants be renamed as `DEFAULT_TILT_RANGE_MAX` and `DEFAULT_TILT_RANGE_MIN`?

Janeece:
@codemonkey800 looks good Jeremy, thank you!

One small thing i noticed (not a blocker, i'd say low severity) - when i apply a min/max value and then reopen the filter to edit, the values i previously applied are not there. I'd expect my values to still be in the fields when I reopen.

manasaV3 (approved):
lgtm!

Issue #322, opened 2024-01-09
Update the API client to return lists instead of generator objects for related objects


jgadling:
The API client currently returns generator objects for related classes (ex: `dataset.runs`) and this is confusing to our users. Update the client to return a list for these relationships instead.
- Make sure we have unit tests for this functionality
- This *may* prove to be difficult because of the way our API client works. Reach out to Utz/Kandarp if it becomes very sticky and we may choose to update the docs instead of the API client.

junxini:
addressed by https://github.com/chanzuckerberg/cryoet-data-portal/pull/412

PR #412, opened 2024-01-24
[breaking] Updates the client to return lists instead of generators for all find() and n:many relationships

jgadling:


manasaV3 (approved):
:shipit:

junxini:
relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/322

Issue #358, opened 2024-01-18
filter on single dataset page - to narrow down runs by Annotation Object Name


junxini:
Add another filter to the single dataset page:

- [ ] title of filter: Annotation Object Name 
- [ ] filter dropdown to list all annotation object names that are included in the dataset (i.e., the filter dropdown values will change depending on what is available for that given dataset)
- [ ] filter type: multi-select
- [ ] filter will match the existing "Object Name" filter functionality found on the browse data page

junxini:
@codemonkey800 i've added this to the sprint queue since it was groomed during backlog refinement. please feel free to move back if you cannot take this on in sprint 2!

PR #417, opened 2024-01-29
annotated object name filter

codemonkey800:
Implements #358
Depends on #407

- Creates reusable annotated object name filter component for use on the dataset and run filters
- Adds name filter to the run filter

## Demo

https://dev-filter-test.cryoet.dev.si.czi.technology/browse-data/datasets

image


kne42:
any chance you can spin up a dev server for ppl to test with?

Janeece:
@codemonkey800 Only thing I noticed is happening with the tilt quality too — filter tag is showing beside the filter dropdown instead of below
Screenshot 2024-02-01 at 12 04 50 PM

Issue #343, opened 2024-01-18
There is a refresh hapenning while navigating from page to page and the loader is flashed


lauramarcos:
Steps to Reproduce

1. Go to https://cryoetdataportal.czscience.com/
2. Click on Browse Data
3. Click on a Dataset Name
4. Click on "Back to Results"

**Actual Results**
Loader is displayed for a moment

**Evidence**

https://github.com/chanzuckerberg/cryoet-data-portal/assets/63799148/9803268a-0c6b-45ec-b2ae-4fbf1ab2bcc5

junxini:
i think there was a PR that fixed this. could you help verify @lauramarcos?

lauramarcos:
This is no longer showing up.. closing bug.. thank you!

PR #428, opened 2024-01-31
fix refresh bug on initial load

codemonkey800:
#343

Fixes issue with loader flashing momentarily on initial load. The root cause was we were calling `setSearchParams()` on initial load, which forces a refetch and re-render


frontend/packages/data-portal/app/components/BrowseData/BrowseDataSearch.tsx
```diff
@@ -17,8 +17,15 @@ export function BrowseDataSearch() {
   const [query, setQuery] = useState(searchParams.get('search') ?? '')
 
   // If the user hasn't typed in a key for 500ms, then update the search params.
+  const initialLoadRef = useRef(true)
```
richaagarwal, line 20: Does this get initialized/re-initialized any time a user is navigating to search results (including clicking "Back to Results")?

richaagarwal (approved):

PR #452, opened 2024-02-06
Add missing documentation for TiltSeries fields.

jgadling:
I added an alphabetic sort (a no-op) to most of the model attributes, which makes it easier to compare the docstrings to the properties of these classes.

The functional change is that I added some missing fields to the TiltSeries docstring.


manasaV3 (approved):
🚢

Issue #440, opened 2024-02-03
"View Tomogram" as a Plausible custom goal


bnelson-czi:
Right now, we track clicks to View Tomogram via [Outbound Links in Plausible](https://plausible.io/share/cryoetdataportal.czscience.com?auth=jf9cJeBWiRymfylFkd096&goal=Outbound+Link%3A+Click). However, Plausible breaks these clicks down by unique tomograms, so it's incredibly granular and difficult to aggregate. We would like a more general sense of how many times visitors click to view tomogram, which we should be able to track as a Plausible goal with the following properties:
- dataset ID
- run ID
- whether click happened at the dataset page level or run page level

Link to original comment for additional context: https://github.com/chanzuckerberg/cryoet-data-portal/issues/109#issuecomment-1830396496

bnelson-czi:
I saw [the PR](https://github.com/chanzuckerberg/cryoet-data-portal/pull/453) got merged and tested my own actions in prod.
- First tested View Tomogram on two runs (ids: 12371, 12379) on [this dataset page](https://cryoetdataportal.czscience.com/datasets/10141). Plausible recorded both.
- I clicked to a [run page](https://cryoetdataportal.czscience.com/runs/12394) and View Tomogram. I did not see this event in Plausible.
- Went back to the dataset page and clicked to View Tomogram on other runs. I don't see these events either.

So Plausible was recording single dataset View Tomogram events at first, but it stopped working after I tried it at the run level. Can anyone replicate this?

PR #453, opened 2024-02-06
single dataset view tomogram goal

codemonkey800:
#440

- Create reusable component `ViewTomogramButton` that shares the plausible tracking logic in one component
- Refactor run header and run table to use new view tomogram button

## Demo

image


kne42 (approved):
nice!

PR #467, opened 2024-02-09
Show frontend URL in workflow run summary

codemonkey800:
Instead of digging into the run output, the URL will be conveniently be available in the run summary

## Demo

https://github.com/chanzuckerberg/cryoet-data-portal/actions/runs/7837731462

image


manasaV3 (approved):
🚢

Issue #424, opened 2024-01-30
GitHub integration


codemonkey800:
Build out the GitHub workflow to automatically run E2E tests when code is merged to main

PR #473, opened 2024-02-13
E2E tests workflow

codemonkey800:
#424

- Creates reusuable workflow for running E2E tests
- Adds workflow after staging deployments
- Fixes issue with E2E tests being flaky on initial load

### DIFF
--- a/.github/workflows/staging-deploy.yml
+++ b/.github/workflows/staging-deploy.yml
@@ -6,7 +6,6 @@ on:
       - main
     paths:
       - 'frontend/**'
-      - 'frontend/.happy/**'
 
 # https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
 permissions:


.github/workflows/staging-deploy.yml
```diff
@@ -3,7 +3,9 @@ name: Deploy to staging
 on:
   push:
     branches:
-      - "main"
+      - main
+    paths:
+      - 'frontend/**'
```
manasaV3, line 8: We should also add happy path here.

.github/workflows/frontend-e2e-tests.yml
```diff
@@ -0,0 +1,80 @@
+on:
+  workflow_call:
+    inputs:
```
manasaV3, line 3: Reusable workflow components for the win! 🎉

.github/workflows/frontend-e2e-tests.yml
```diff
@@ -0,0 +1,80 @@
...
+
+    secrets:
+      E2E_CONFIG:
+        required: true
+
+jobs:
+  frontend-e2e-tests:
+    strategy:
+      matrix:
+        browser: [chrome]
```
manasaV3, line 16: Curious about why we only test for Chrome?

.github/workflows/staging-deploy.yml
manasaV3: Forgot, we moved the `.happy` folder inside `frontend`.

frontend/packages/data-portal/e2e/filters/utils.ts
```diff
@@ -64,3 +64,19 @@ export async function validateTable({
...
   await validateRows(page)
 }
+
+/**
+ * When loading the page, we need to wait a bit after so that the SDS components
+ * have time to become interactive. Without the timeout, the tests become more
+ * flaky and occasionally fail. For example, the filter dropdowns sometimes
+ * don't open when clicked on because the playwright browser starts clicking on
+ * it too fast while the JavaScript is still loading and hydrating.
+ */
```
kne42, line 74: weird. how much slower does this make our tests?

kne42 (approved):

.github/workflows/staging-deploy.yml
```diff
@@ -3,7 +3,10 @@ name: Deploy to staging
...
+      - 'frontend/.happy/**'
```
manasaV3, line 9: Sorry, I marked my earlier comment as resolved. We don't need this, as it should be covered by the `frontend/**` line.

manasaV3 (approved):
🚀

Issue #432, opened 2024-02-01
Annotation table updates (MVP without dividers)


junxini:

manasaV3:
As a part of this revamp, we should also group the records by object shape type such that there is only one record for Segmentation Mask as opposed to the Zarr and MRC records that we currently display.

PR #481, opened 2024-02-20
feat: annotation table updates

codemonkey800:
#432

- Adds rudimentary feature flagging module that can be controlled via code or query parameter
- Updates the annotation table to match the updated design
  - Implements but hides method type column

## Demo

https://dev-jeremy-anno-table.cryoet.dev.si.czi.technology/runs/242

image

### With Method Type

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
@@ -317,17 +317,7 @@ export function AnnotationTable() {
   ])
 
   const annotations = useMemo<Annotation[]>(
-    () =>
-      run.annotation_table.flatMap((data) =>
-        data.annotations.map((annotation) => ({
-          ...annotation,
-          authors: annotation.authors.concat(
-            { name: 'Foo Bar' },
-            { name: 'Bar Foo' },
-            { name: 'Jeremy Asuncion' },
-          ),
-        })),
-      ),
+    () => run.annotation_table.flatMap((data) => data.annotations),
     [run.annotation_table],
   )
 


frontend/packages/data-portal/app/components/Run/AnnotationObjectTable.tsx
```diff
@@ -35,7 +35,7 @@ export function AnnotationObjectTable() {
         },
         {
           label: t('objectShapeType'),
-          values: [annotation.shape_type],
+          values: [annotation.files[0].shape_type],
```
kne42, line 38: why this change? did the data model itself change?

frontend/packages/data-portal/app/utils/featureFlags.ts
```diff
@@ -0,0 +1,45 @@
...
+export const FEATURE_FLAGS = {
+  methodType: 'dev',
+} as const
+
+export type FeatureFlagKey = keyof typeof FEATURE_FLAGS
+
+const ENABLE_FEATURE_PARAM = 'enable-feature'
+const DISABLE_FEATURE_PARAM = 'disable-feature'
+
+export function getFeatureFlag({
```
kne42, line 16: this seems more like a user setting than anything – would we be accomplishing this stuff with cookies in the future?

kne42 (approved):
LGTM, had some questions

Janeece:
@codemonkey800 sorry I'm late to the party on this one, i have one comment — the table cell link for the method type should not be the minimal button, should just be a normal text link. 
Screenshot 2024-03-07 at 4 57 19 PM

Not a critical-blocker but it's over emphasized + confusing in how it competes with the action buttons on the right. Would love to see this updated if you have the time, otherwise logged as a bug to fix later.

frontend/packages/data-portal/app/utils/featureFlags.ts
```diff
@@ -0,0 +1,45 @@
+import { useSearchParams } from '@remix-run/react'
+
+import { useEnvironment } from 'app/context/Environment.context'
+
+export type FeatureFlagEnvironment = typeof process.env.ENV
+
+export const FEATURE_FLAGS = {
+  methodType: 'dev',
```
manasaV3, line 8: We should update the data structure so that we can enable a feature flag for multiple environments.

PR #508, opened 2024-02-26
chore: fix release workflow

codemonkey800:
#437

- Rename package name to `cryoet-data-portal-web` to be clearer it's a frontend tag
- Adds workflow to generate frontend release PR for every push to main
  - Any existing PRs will edited to include the updated changelog
- Refactored the prod workflow
  - Fix issue with workflow not detecting previous release tags
  - Able to tag a release after PR is merged
  - Adds a comment to the PR with a link to the tag the PR was released in
  - Add workflow dispatch option to deploy to Terraform and skip PR merging + tagging 

## Demos

All example runs, PRs, and tags can be found in my fork:
https://github.com/codemonkey800/cryoet-data-portal

image

image

### DIFF
--- a/.github/workflows/prod-deploy.yml
+++ b/.github/workflows/prod-deploy.yml
@@ -1,7 +1,10 @@
-name: Frontend Production Release
+name: Deploy to prod
 
 on:
   workflow_dispatch:
+  schedule:
+    # Weekdays Mon-Fri @ 18:00 UTC (10AM PST [winter] / 11AM PDT [summer])
+    - cron: '0 18 * * 1-5'
 
 permissions:
   contents: write
@@ -9,6 +12,15 @@ permissions:
   pull-requests: write
 
 jobs:
+  run-staging-e2e-test:
+    name: Run e2e test on staging
+    uses: chanzuckerberg/cryoet-data-portal/.github/workflows/frontend-e2e-tests.yml@main
+    secrets: inherit
+    # The secret IS_DEPLOY_ALLOWED is used to manually gate the prod deploys
+    if: ${{ secrets.IS_DEPLOY_ALLOWED == 'true' }}
+    with:
+      environment: staging
+
   find-frontend-release-pr:
     concurrency:
       group: create-frontend-release-${{ github.ref }}
@@ -19,9 +31,8 @@ jobs:
 
     name: find release PR
     runs-on: ubuntu-latest
-
     if: github.repository == 'chanzuckerberg/cryoet-data-portal'
-
+    needs: run-staging-e2e-test
     steps:
       - name: find release PR
         uses: actions/github-script@v7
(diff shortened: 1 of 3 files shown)


.github/workflows/prod-deploy.yml
```diff
@@ -1,67 +1,136 @@
...
-      prs_created: ${{ steps.release.outputs.prs_created }}
+      pr: ${{ steps.find-release-pr.outputs.result }}
 
-    name: generate frontend changelog
+    name: find release PR
     runs-on: ubuntu-latest
-    if: github.repository == 'chanzuckerberg/cryoet-data-portal'
+
+    if: |
+      github.repository == 'codemonkey800/cryoet-data-portal' &&
```
manasaV3, line 28: I believe this should be `chanzuckerberg/cryoet-data-portal`. 🤓

.github/workflows/prod-deploy.yml
```diff
@@ -1,67 +1,136 @@
+name: Frontend Production Release
+
 on:
   workflow_dispatch:
+    inputs:
+      skip-tagging:
+        type: boolean
+        description: Skip merging and tagging a release.
```
manasaV3, line 8: Is this needed?

frontend/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "cryoet-data-portal",
+  "name": "cryoet-data-portal-web",
```
manasaV3, line 2: 🔥

.github/workflows/prod-deploy.yml
manasaV3: I believe the cause for why we ran into issues with needing to skip tagging has been addressed as a part of this PR. Do you think we could still run into a similar situation?

.github/workflows/prod-deploy.yml
```diff
@@ -1,67 +1,136 @@
...
+      - name: get tag name
+        id: get-tag-name
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const pr = ${{ needs.find-frontend-release-pr.outputs.pr }}
+            const version = pr.title.split(' ').at(-1)
+
+            return {
+              version,
```
manasaV3, line 112: Are we using the version from this anywhere? If not, would it be cleaner just to have it return the tag, so we don't have to parse the json wherever we are trying to get the tag?

manasaV3:
We should also have a way to rerun the workflow for the case where a version was tagged successfully, but requires redeployment because the prod deploy failed. 🙈

.github/workflows/prod-deploy.yml
manasaV3: I believe it won't solve for it, because we are querying for `label:"autorelease: pending"` in the initial step. But, if it fails on deployment, it wouldn't have that label. It would have the label `autorelease: tagged`.

manasaV3 (approved):
:shipit:

PR #514, opened 2024-02-29
fix: fix deploy var

codemonkey800:
Fixes issue with prod deploy due to secrets not being available for job conditionals: https://github.com/actions/runner/issues/520

This switches to using a repository variable which should be appropriate since the value can only be `true` or `false`


manasaV3 (approved):

manasaV3 (commented):
I believe this would need the gating variable to be defined in a different environment than what it is gating. Not blocking on that, but we should fix that in near future. 

From async: we are using global variables for `IS_DEPLOY_ALLOWED`, so this concern is no longer valid.

Issue #389, opened 2024-01-22
Run Name is cut short in the Runs Table - Single Dataset Page 


junxini:
Bug description: When run names are long, the title gets cutoff because the column width is too narrow

Expected Result: wider column to decrease likelihood of the run name getting cutoff

![image](https://github.com/chanzuckerberg/cryoet-data-portal/assets/98626158/5b7b6b83-8335-4c63-95b7-0a60a7903b09)

Link: https://cryoetdataportal.czscience.com/datasets/10004

See Janeece's comment below for design specs

Janeece:
@junxini @manasaV3 @codemonkey800 @kne42 

Here is the updated spec for the runs table spacing (just frame 1.10 Min/Max sizing, it's marked for this ticket/ready for dev):
https://www.figma.com/file/0UT1iyRVRkIdH4qLfh8JDd/3---Single-Dataset?type=design&node-id=94%3A9403&mode=dev

- The pink text with "update" in bold at the end are the changes from the original spec. 
  - Run Column: 
     - Remove max width
     - changed min width to 400px 
     - defined Run name text wrap up to max content height (100px), then overflow
  - Button Column: 
     - Add 300px max width
- These changes favor more space for the run name to minimize instances of cut-off. Some run tables will have a large gap of space because they are short, but I think as a stop gap to a better overflow solution, its better to have too much space than important info being cut off. 

Let me know if there are any questions or concerns!

<img width="1481" alt="Screenshot 2024-01-22 at 4 03 15 PM" src="https://github.com/chanzuckerberg/cryoet-data-portal/assets/63013360/9f2f2cbb-2694-465c-a633-3f524b103a84">

PR #551, opened 2024-03-07
fix: table column sizing

codemonkey800:
#389

- Fixes the table column sizing for min and max widths
- This updates it so that the table header also gets the same min and max widths as the column
- There was an issue where the tables were not filling the remaining space
  - This is because the min and max widths need to be defined for the header as well

### DIFF
--- a/frontend/packages/data-portal/app/components/Table/CellHeader.tsx
+++ b/frontend/packages/data-portal/app/components/Table/CellHeader.tsx
@@ -17,6 +17,7 @@ export function CellHeader({
   showSort = false,
   tooltip,
   width: columnWidth,
+  children,
   ...props
 }: Omit<
   CellHeaderProps,
@@ -48,6 +49,8 @@ export function CellHeader({
         minWidth: columnWidth?.min,
       }}
       hideSortIcon={!showSort}
-    />
+    >
+      <span className="line-clamp-1">{children}</span>
+    </SDSCellHeader>
   )
 }


kne42:
> There was an issue where the tables were not filling the remaining space
> This is because the min and max widths need to be defined for the header as well
oh, very interesting

could you spin up a dev server for the design folks to check out?

Janeece:
@codemonkey800 Here's my comments -- looks good and none of these are blockers, thank you!

### All datasets table:
looks good, behaves as expected

### Runs table (single dataset page):
This looks way better & achieves the goal. 
I'm wondering if we can make it so that the "tilt series quality score" header & tags display a little more nicely:
1.  Would be nice if "tilt series quality score" header didn't need to wrap at larger screen widths (looks silly when theres so much space)
2. would be nice if min-width on ts quality column prevented text in the text in the tags from wrapping. 
3. would prefer column overflow to be hidden behind ellipses instead of wrapping (tooltip gives user the full name on hover)

I'm open to any suggestions on how to achieve those refinements. I'm thinking we could tweak some of the widths a little:
- reduce the Annotated Object column max-width from 500px to 400px 
- reduce "view tomogram" button column max-width from 300px to 200px
- increase "tilt series quality score" min-with so that the excellent tag doesnt wrap (about 110px?)

### Annotations table (single runs page):
- overall looks as expected
- The columns don't seem to have the max-widths as defined in most [recent spec](https://www.figma.com/file/gwX2PjfGPfMrbEWMMJq6rW/H1-2024---Annotations-Table-Update?type=design&node-id=639%3A177561&mode=design&t=KUrNgFNki2L9VC9F-1) - we will want these to have more space for authors in the future. 
- table width behaves differently than the others (no max width for content) - I'm not too worried about this now, but eventually we'll want to consider how to treat max-table widths consistently (after the info in the table becomes a bit more stable).

kne42 (approved):

Janeece:
@codemonkey800 LGTM! Seems to fix the core issue and overall is much improved! no worries about the ongoing investigation in to the table max, not a blocker for me. Thanks for checking back in.

Issue #456, opened 2024-02-07
Tilt range filter and metadata update


junxini:
**Update the tilt range filter to clarify it's referring to the range and not the axis:**

- [ ] Filter description: leave blank (i.e., remove "Add min/max angle to define filter range")
- [ ] Filter subtext: change location to below input fields, no copy changes
- [ ] input fields: remove the equation sign, label as "Min Tilt Range:" and "Max Tilt Range:" 
- [ ] Location of change: 
  - Browse all filter panel
  - Dataset filter panel

Link to design: https://www.figma.com/file/oUtMvkDdADObFmvOJVVy9G/Phase-2-Deprioritized-Designs-%26-Explorations?node-id=289%3A1505&mode=dev

**Update the metadata to specify it's referring to the range:**

- [ ] Update Tilt Range field to show "X°(Y° to Z°)"
  - X = the difference between Y and Z
  - concrete example: "98° (-40° to 58°)”
- [ ] Location of change: 
  - Single Run page, header area
  - Single Run metadata panel
 
Link to design: N.A

PR #552, opened 2024-03-07
chore: update tilt range label

codemonkey800:
#456

- Adds reusable util for creating tilt range label
- Updates the tilt range label for the tilt series table and run header table

## Demo

image

image


kne42 (approved):

manasaV3 (approved):
🚢

Issue #530, opened 2024-03-04
Handle runs that have only tomogram or only tiltseries


manasaV3:
Currently we throw error for runs that have a tiltseries but no tomogram. Example: https://cryoetdataportal.czscience.com/runs/958

manasaV3:
Fix by hiding the tomogram metadata panel in the more info.

junxini:
this ticket is specific fixing the error by hiding the section the tomogram section of the metadata when tomogram is not available. design will work on edge case tracked in #547

PR #559, opened 2024-03-07
fix: no tomogram error

codemonkey800:
#530

Fixes issue with frontend throwing an application error when opening the metadata drawer when there are no tomograms in the tilt series

### Before

image

### After

Hides tomogram metadata panel when undefined 

image


manasaV3 (approved):

kne42 (approved):
yay it was a simple fix!

Issue #436, opened 2024-02-01
Method type tool tip on annotation table


junxini:
tooltip within the Annotation Table:
- [ ] tool tip when hovering over "Method Type" column header and the 3 possible values within that column - Manual, Automated, Hybrid ([see design here](https://www.figma.com/file/gwX2PjfGPfMrbEWMMJq6rW/H1-2024---Annotations-Table-Update?type=design&node-id=639%3A178860&mode=dev))

dependent on #464

dgmccart:
[Copy is linked in this issue](https://github.com/chanzuckerberg/cryoet-data-portal/issues/464) @kne42

PR #568, opened 2024-03-09
feat: implement method type tooltips

kne42:
#436 

Screenshot 2024-03-08 at 7 40 29 PM
Screenshot 2024-03-08 at 7 40 39 PM
Screenshot 2024-03-08 at 7 41 06 PM
Screenshot 2024-03-08 at 7 41 31 PM

note: I am aware that the tooltip boxes don't align exactly like the design. Getting them centred is not a trivial fix, so if we do want to get around to it I suggest capturing it in a P1 issue. However, I can shift them a bit left in the meantime.


Janeece (approved):
@kne42 LGTM, thank you and apologies about the delay!

Issue #488, opened 2024-02-21
Additional Filters on Single Dataset Page


junxini:
Add 2 filters to the single dataset page so that the filtering experience is consistent with the new additions on the single run page:
- Ground Truth Annotation
- Object Shape Type

Link to design: https://www.figma.com/file/q6Z394Xy6wUmaXQ9YFjZH7/Kevin---2023-Spillover-Work?type=design&node-id=2118-11276&mode=design&t=gxrH9lSYcXlkEzFC-0

PR #573, opened 2024-03-13
feat: add additional filters to single dataset page

kne42:
#488 

- adds ground truth annotation filter to runs table
- adds object shape type filter to runs table

Screenshot 2024-03-12 at 6 49 30 PM


kne42:
live demo: https://dev-kira-1.cryoet.dev.si.czi.technology/

Janeece:
I'll have @kev-zunshiwang review this PR since he worked on this project.

kev-zunshiwang:
LGTM!

There's missing spaces between the words ie. SegmentationMask
**Question:** is this only specific to this work or might touch on a couple of areas?  
**Priority:** If only applies to this work ill assign a P1, if it touches on a couple of areas then I can flag to Jun for a backlog ticket.
Missing the "Point" option in demo (maybe the work is not done yet)
**Priority:** this would be a blocker for release

Thanks Kira!

kne42:
> There's missing spaces between the words ie. SegmentationMask
> **Question:** is this only specific to this work or might touch on a couple of areas?

This is an issue across all filters. These names are autopopulated from the backend so either we could implement a fix there or perhaps if there are a discrete number of options we can have a mapping on the frontend.

> Missing the "Point" option in demo (maybe the work is not done yet)

These options are autopopulated from the list of annotation files available. Are you saying that there was an annotation of shape type "Point" in the table but the list of options didn't have it available for filtering?

kev-zunshiwang:
There's missing spaces between the words ie. SegmentationMask
Question: is this only specific to this work or might touch on a couple of areas?

This is an issue across all filters. These names are autopopulated from the backend so either we could implement a fix there or perhaps if there are a discrete number of options we can have a mapping on the frontend.

@junxini do we have a backlog ticket for this?

kev-zunshiwang:
> Missing the "Point" option in demo (maybe the work is not done yet)
> 
> These options are autopopulated from the list of annotation files available. Are you saying that there was an annotation of shape type "Point" in the table but the list of options didn't have it available for filtering?

I would expect all options to display, and the ones not available for this specific dataset, the option(s) are disabled.
Sounds like we don't do this, is this correct? @kne42

junxini:
> @junxini do we have a backlog ticket for this?
we don't have an existing ticket for this yet. it will be a part of the metadata refinement work

PR #576, opened 2024-03-13
chore: support multiple envs for feature flags

codemonkey800:
Adds support for enabling a feature flag for multiple environments

### DIFF
--- a/frontend/packages/data-portal/app/utils/featureFlags.ts
+++ b/frontend/packages/data-portal/app/utils/featureFlags.ts
@@ -4,11 +4,11 @@ import { useEnvironment } from 'app/context/Environment.context'
 
 export type FeatureFlagEnvironment = typeof process.env.ENV
 
-export const FEATURE_FLAGS = {
-  methodType: 'dev',
-} as const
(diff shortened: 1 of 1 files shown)


frontend/packages/data-portal/app/utils/featureFlags.ts
```diff
@@ -4,11 +4,11 @@ import { useEnvironment } from 'app/context/Environment.context'
...
 export type FeatureFlagEnvironment = typeof process.env.ENV
 
-export const FEATURE_FLAGS = {
-  methodType: 'dev',
-} as const
+export type FeatureFlagKey = 'methodType'
 
-export type FeatureFlagKey = keyof typeof FEATURE_FLAGS
+export const FEATURE_FLAGS: Record<FeatureFlagKey, FeatureFlagEnvironment[]> = {
+  methodType: ['local', 'dev'],
```
manasaV3, line 10: Not a blocker: For local environments, should we always default to true or have an env override so that we don't accidentally update the configuration from local testing? 

We can always catch this in the code reviews, so it is not a blocker.

manasaV3 (approved):

Issue #572, opened 2024-03-12
Remove "+n more" authors button within tables


junxini:
**Context:** To streamline user navigation and encourage click-through to the next level of data, we are removing redundant buttons. the "+n more" button opens up the metadata panel which is already covered in the "more info" button, hence not needed.

**Tasks**
- remove the "+n more" button next to the author names on the following pages:
  - browse all page
  - single run page 

**Dependency**
- #524 needs to be implemented before we can remove the button 

**Link to design**: https://www.figma.com/file/tKMoeyLbzQDLtlyrEzs9uX/Kevin---Data-Discovery-Improvements-V1?type=design&node-id=4585-104623&mode=design&t=IiLun97AOqy4hyEC-0

Issue #524, opened 2024-03-01
Update styling of author name within tables


junxini:
Change the author listing so that primary and corresponding authors are displayed in the table by default. This way,  users don't have to click into it to see this information.
- the format will follow: "primary author, primary author, ..., corresponding author, corresponding author". The ... is meant to signify other authors that are hidden by default. In cases where there are no authors who aren't primary or corresponding, we should not show the ellipsis.  
- authors are separated by commas, no bold or envelope styling, all text in gray
- table height will vary depending on the length of the primary+corresponding author length
- this change will be reflected in the Annotation table and the Browse all table

Link to design: https://www.figma.com/file/tKMoeyLbzQDLtlyrEzs9uX/Kevin---Data-Discovery-Improvements-V1?type=design&node-id=4328-16358&mode=design&t=60MdOt6Cns3EFZzy-0

PR #580, opened 2024-03-14
style: update author name styling in tables

kne42:
#524, #572
- removes "show more" for author lists in tables
- collapses non-primary, non-corresponding authors with an ellipsis
- top aligns dataset thumbnail when overflowing in table view

please note that there may be some odd-appearing author lists, but this is due to data that needs to be fixed, since every dataset is supposed to have at least one primary and one corresponding author


kne42:
@kev-zunshiwang this is ready for re-review

kev-zunshiwang:
LGTM!

Issue #628, opened 2024-04-05
download_all_annotations() downloads metadata unrelated to specified shape


kandarpksk:
In the code below shared by a participant during a usability study session, they are trying to download only point annotations (say), but metadata related to segmentation masks is also downloaded, causing confusion. Similar, if they try to download only segmentation masks (last line of code), metadata related to point annotations is also downloaded.

```
tomos = portal.Tomogram.find(client, [portal.Tomogram.tomogram_voxel_spacing.run.dataset.id == "10000"])
for t in tomos:
	# uncomment either of the below lines at a time
	t.download_all_annotations(shape="Point", format="ndjson")
	# t.download_all_annotations(shape="SegmentationMask", format="mrc")
```

PR #639, opened 2024-04-12
fix: Don't dowload too many annotation metadata files

jgadling:
This is the fix for #628 and prevents `download_all_annotations()` from downloading metadata files for annotations that are filtered out by shape or format filters.


kandarpksk (approved):
LGTM!

manasaV3 (approved):
🚢

Issue #626, opened 2024-04-05
Comparing annotation object to string returns results, but should not


kandarpksk:
In the code written below by a participant during a usability study session, annotation objects are compared to a string, which doesn't make sense. However, many matches are found, which is a bug – I expect the comparison to error out due to a type mismatch, or at least fail. cc: @uermel

```
import cryoet_data_portal as portal
client = portal.Client()
runs = portal.Run.find(client, query_filters=[portal.Run.tomogram_voxel_spacings.annotations == "membrane"])
for r in runs:
	print(r.dataset_id)
```

manasaV3:
We should update the copy on the download dialog box to use `sync` instead of `cp`.

junxini:
@jgadling we forgot to point this in the meeting earlier :( Is this something you could help point async?

PR #642, opened 2024-04-12
fix: Raise exception when attempting to filter on related objects.

jgadling:
Fix for #626 , raise an exception when filtering on invalid fields.

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
@@ -21,7 +21,7 @@ class GQLExpression:
                 fieldname = [fieldname]
 
             raise Exception(
-                f"\"{'.'.join(fieldname)}\" is an object and can't be filtered directly. Please filter on one of its scalar attributes instead: {', '.join(scalar_fields)}"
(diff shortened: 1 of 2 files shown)


client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
```diff
@@ -14,6 +14,15 @@ def __init__(self, field: "GQLField", operator, value: Optional[Any]):
...
     def to_gql(self) -> Dict[str, Any]:
         fieldname = self.field.to_gql()
+        if self.field.is_relationship():
+            related_cls = self.field.get_related_class()
+            scalar_fields = related_cls._get_scalar_fields()
+            if isinstance(fieldname, str):
+                fieldname = [fieldname]
+
+            raise Exception(
+                f"\"{'.'.join(fieldname)}\" is an object and can't be filtered directly. Please filter on one of its scalar attributes instead: {', '.join(scalar_fields)}",
```
kandarpksk, line 24: Do we want to say "filtered directly" here or "compared directly"?

jgadling: I updated the message!

manasaV3 (approved):
🚢

PR #643, opened 2024-04-15
fix: List all shapes available for a given annotation.

### LINKED ISSUE #638: Not all annotation shape types are displayed in the annotation table on "Single Run" page
The expected behavior is: If an annotation has multiple files for the same `shape_type` there is only one row in the Annotations table, but if there are multiple values for `shape_type` for one annotation, there is a row for each `shape_type`. 

Example: 
[Run 277](https://cryoetdataportal.czscience.com/runs/277), Annotation 2395

This Annotation has 3 associated files (2 for `shape_type`=="SegmentationMask", 1 for `shape_type`=="InstanceSegmentation"). 

The table should display: 
| Annotation ID | Deposition Date | Object Name | Object Shape Type | Precision | Recall |
| -- | -- | -- | -- | -- | -- |
| 2395  | 2024-02-20  | Membrane | InstanceSegmentation |  N/A |  N/A |
| 2395  | 2024-02-20  | Membrane | SegmentationMask |  N/A |  N/A |

Currently it only displays: 
| Annotation ID | Deposition Date | Object Name | Object Shape Type | Precision | Recall |
| -- | -- | -- | -- | -- | -- |
| 2395  | 2024-02-20  | Membrane | InstanceSegmentation | N/A |  N/A |

The API client shows that all files are stored correctly in the database:

```python
from cryoet_data_portal import Client, AnnotationFile

client = Client()
anno_files = AnnotationFile.find(client, [AnnotationFile.annotation.tomogram_voxel_spacing.run.id == 277])

[f"{af.annotation.id} - {af.annotation.object_name}: {af.shape_type} {af.format}" for af in anno_files]
```

> ['2395 - membrane: InstanceSegmentation ndjson',
>  '2395 - membrane: SegmentationMask zarr',
>  '2395 - membrane: SegmentationMask mrc',
>  '1292 - Ribosome, 80 S: Point ndjson',
>  '15288 - cytosolic ribosome: Point ndjson']

jgadling:
Fixes #638 

An annotation may contain files that represent multiple different annotation shapes, we should list all unique shape types available for an annotation. This produces a result like so:

Screenshot 2024-04-15 at 12 00 39 PM

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
@@ -259,7 +259,7 @@ export function AnnotationTable() {
 
         cell: ({ getValue }) => (
           <TableCell width={AnnotationTableWidths.files}>
-            {getValue().at(0)?.shape_type ?? '--'}
+            {[...new Set(getValue().map((item) => item.shape_type))].join(', ')}
           </TableCell>
         ),
       }),
--- a/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
@@ -259,7 +259,9 @@ export function AnnotationTable() {
 
         cell: ({ getValue }) => (
           <TableCell width={AnnotationTableWidths.files}>
-            {[...new Set(getValue().map((item) => item.shape_type))].join(', ') ?? '--'}
+            {[...new Set(getValue().map((item) => item.shape_type))].join(
+              ', ',
+            ) ?? '--'}
           </TableCell>
         ),
       }),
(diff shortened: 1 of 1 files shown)


manasaV3:
iirc, according to the original design, we should have an individual record for each shape type.

jgadling:
> iirc, according to the original design, we should have an individual record for each shape type.
Our schema design supports multiple shapes per annotation, probably intentionally?

manasaV3:
> > iirc, according to the original design, we should have an individual record for each shape type.

>  Our schema design supports multiple shapes per annotation, probably intentionally?

Yes, that is correct. We should be able to support a single annotation with multiple different shapes.

jgadling:
> > > iirc, according to the original design, we should have an individual record for each shape type.
> 
> > Our schema design supports multiple shapes per annotation, probably intentionally?
> 
> Yes, that is correct. We should be able to support a single annotation with multiple different shapes.

You're saying the UI needs to draw a whole new row for each shape instead of listing multiple shapes on one row?

Issue #635, opened 2024-04-11
Primary Author/Annotator missing in Annotation Table Rows


uermel:
The primary author's/annotator's name is missing in the Annotation table (but not the metadata panel). 

See example below.

<img width="319" alt="Screenshot 2024-04-11 at 3 26 39 PM" src="https://github.com/chanzuckerberg/cryoet-data-portal/assets/6641113/acb7ed34-2edb-4c90-965b-3f0cce613a2f">

PR #649, opened 2024-04-16
fix: annotation table primary author sort

codemonkey800:
#635

Fixes issue with primary author not being visible on the annotation table

## Demo

image


frontend/packages/data-portal/app/graphql/getRunById.server.ts
```diff
@@ -149,7 +149,7 @@ const GET_RUN_BY_ID_QUERY = gql(`
             shape_type
           }
 
-          authors(order_by: { primary_annotator_status: desc }) {
```
manasaV3, line 152: Can we use author_list_order here?

manasaV3 (approved):

PR #591, opened 2024-03-20
feat: download single tomogram modal updates

codemonkey800:
#439

- Fixes issue with select not auto closing
- Updates rendering for the active tomogram resolution
- Add file format dropdown
- Add code snippet

## Demos

https://dev-jeremy-dl-modal.cryoet.dev.si.czi.technology

<details>
<summary>Expand to see all demos</summary>

### Configure Download

image

#### `.mrc` Direct Download

image

#### `.mrc` Curl

image

#### `.mrc` AWS

image

#### `.mrc` API

image

#### `.zarr` AWS

image

#### `.zarr` API

image

</details>

PR #589, opened 2024-03-19
feat: single annotation dialog step 2

codemonkey800:
#450

Implements step 2 for the download annotation dialog

- Refactor to pass `activeAnnotation` instead of annotation map
- Adds `react-query` for client side fetching
- Refactor file size fetching to be more efficient
  - instead of fetching all file sizes for the page, fetch on demand when user views file size in the dialog
- Fix navigation for step 2

## Demos

https://dev-jeremy-dl-modal.cryoet.dev.si.czi.technology

<details>
<summary>Expand to see demos</summary>

### Configure Download

image

#### `.mrc` Direct Download

image

#### `.mrc` Curl

image

#### `.mrc` AWS

image

#### `.mrc` API

image

#### `.ndjson` Direct Download

image

#### `.ndjson` Curl

image

#### `.ndjson` AWS

image

#### `.ndjson` API

image

#### `.zarr` AWS

image

#### `.zarr` API

image

</details>

PR #665, opened 2024-04-18
ci: fix download modal e2e tests

codemonkey800:
Fixes E2E tests for the download modal after changes to the download modal added the file format parameter in #591 and #589


manasaV3 (approved):
:shipit:

Issue #627, opened 2024-04-05
Downloading all annotations via the AWS CLI fails


kandarpksk:
A participant mentioned during a usability study session that the AWS CLI download command for all annotations in [this run](https://cryoetdataportal.czscience.com/runs/242?prev=%2Fdatasets%2F10000&download-step=download&download-config=all-annotations&download-tab=aws) errors out. It's not known how widespread this issue is, or if it's specific to this run.

> **$ aws s3 --no-sign-request cp s3://cryoet-data-portal-public/10000/TS_027/Tomograms/VoxelSpacing13.480/Annotations Annotations**
> fatal error: An error occurred (404) when calling the HeadObject operation: Key "10000/TS_027/Tomograms/VoxelSpacing13.480/Annotations" does not exist

manasaV3:
We should update the copy on the download dialog box to use `sync` instead of `cp` for downloading all Annotations.

manasaV3:
We should also update the copy of the download dataset dialog box's `via AWS CLI` section to use `aws s3 sync`.

PR #671, opened 2024-04-19
fix: use aws sync for dataset or all annotations download dialogs

codemonkey800:
#627

## Demos

image

image

image

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Download/AWSDownloadTab.tsx
+++ b/frontend/packages/data-portal/app/components/Download/AWSDownloadTab.tsx
@@ -1,4 +1,5 @@
 import { useLocation } from '@remix-run/react'
+import { match, P } from 'ts-pattern'
 
 import { CopyBox } from 'app/components/CopyBox'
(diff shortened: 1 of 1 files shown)


frontend/packages/data-portal/app/components/Download/AWSDownloadTab.tsx
```diff
@@ -1,22 +1,41 @@
...
 export function AWSDownloadTab() {
   const { t } = useI18n()
   const { s3Path } = useDownloadModalContext()
   const { logPlausibleCopyEvent } = useLogPlausibleCopyEvent()
+  const location = useLocation()
+  const { downloadConfig } = useDownloadModalQueryParamState()
+
+  const s3Command =
+    location.pathname.includes('/datasets') ||
+    (location.pathname.includes('/runs') &&
```
manasaV3, line 33: We should also use sync in the AWS download tab for tomogram when it is zarr.

kne42 (approved):

PR #679, opened 2024-04-27
fix: show all annotation shape types

codemonkey800:
#638

Fixes issue with the annotation table now showing each shape type on a separate row

## Demo

image


manasaV3:
Do we have to update any tests as a part of this?

manasaV3 (approved):
🔥

PR #687, opened 2024-05-01
fix: file format race condition

codemonkey800:
There was an issue with calling `setFileFormat` in a use effect because it would have a race condition with other concurrent code that is also updating the URL parameters. The fix is to remove the effect and update the file format in the same code that is replacing the other URL parameters


manasaV3 (approved):

frontend/packages/data-portal/app/hooks/useDownloadModalQueryParamState.ts
```diff
@@ -196,6 +197,7 @@ export function useDownloadModalQueryParamState() {
         [QueryParams.DownloadConfig]: DownloadConfig.Tomogram,
         [QueryParams.TomogramSampling]: initialTomogramSampling,
         [QueryParams.TomogramProcessing]: initialTomogramProcessing,
+        [QueryParams.FileFormat]: 'mrc',
```
kne42, line 200: just triple-checking: mrc will always be an available option, right?

kne42 (approved):

PR #751, opened 2024-05-23
ci: e2e filter utils

codemonkey800:
Extracts utils from the browse dataset filter tests to be reusable across tests. This creates the following functions:

## `testMultiSelectFilter`

Util for testing a filter uses the multi-select component. It tests selecting the specified test options as well as selecting multiple options. If a test query is provided, it will also test the search box is enabled within the filter

## `testMultiInputFilter`:

Util for testing a filter that opens a filter with one or more input text boxes, such as the author or dataset IDs filter


kne42 (approved):
LGTM!

Issue #756, opened 2024-05-25
Fix E2E test for closing dialog failing on ubuntu chromium


codemonkey800:
There is currently an issue with the E2E tests running in the ubuntu test runner that causes some of the tests to consistently fail only for ubuntu chromium. A workaround fix has been created in #757 that switches the test runner to use `macos-latest`

It's related to the tests for closing the download dialog: 
https://github.com/chanzuckerberg/cryoet-data-portal/actions/runs/9212312157/job/25343916915

The exact error message from the test is:

```
Error: Timed out 5000ms waiting for expect(locator).toBeHidden()

Locator: getByRole('dialog')
Expected: hidden
Received: visible
Call log:
  - expect.toBeHidden with timeout 5000ms
  - waiting for getByRole('dialog')
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
  -   locator resolved to <div role="dialog" aria-labelledby=":R536:" class="Mu…>…</div>
  -   unexpected value "visible"
```

However, inspecting the test video shows that the dialog does in fact disappear:

[e2e-close-dialog-test.webm](https://github.com/chanzuckerberg/cryoet-data-portal/assets/2176050/8b79ebe8-aa17-4773-9ed4-60339b7ff68b)

Issue #743, opened 2024-05-22
Fix broken e2e tests for frontend


manasaV3:

junxini:
@codemonkey800 could you point when you have more info?

PR #757, opened 2024-05-25
ci: fix e2e tests

codemonkey800:
#743

Fixes the E2E tests temporarily by switching to the test runner to use `macos`. The root cause for this is still uncertain, but it's work will be tracked in #756.

This also fixes the archive generation for the playwright report by using the correct path 

## Demo

https://github.com/chanzuckerberg/cryoet-data-portal/actions/runs/9237537451

image

### DIFF
--- a/.github/workflows/frontend-e2e-tests.yml
+++ b/.github/workflows/frontend-e2e-tests.yml
@@ -78,6 +78,6 @@ jobs:
       - uses: actions/upload-artifact@v3
         if: failure()
         with:
-          name: playwright-report
+          name: html-report-${{ matrix.browser }}-${{ matrix.shardCurrent }}-${{ matrix.shardTotal }}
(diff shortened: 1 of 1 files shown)


.github/workflows/frontend-e2e-tests.yml
```diff
@@ -22,7 +22,10 @@ jobs:
       cancel-in-progress: true
 
     name: ${{ matrix.browser }} ${{ matrix.shardCurrent}}/${{  matrix.shardTotal }}
-    runs-on: ubuntu-latest
+    # Using macos runner temporarily because ubuntu runner has issues with
+    # playwright testing for when the download dialog closes
+    # TODO investigate why ubuntu runner tests fail
+    runs-on: macos-latest
```
kne42, line 28: woaaah this is super weird

kne42 (approved):

Issue #538, opened 2024-03-06
Adding end to end tests pagination feature


codemonkey800:
Adds E2E tests for pagination. This should check that the pagination counts are correct and that selecting a new page has returns the correct items in the table

PR #776, opened 2024-06-04
ci: pagination e2e

codemonkey800:
#538

Adds pagination E2E tests for the browse datasets and single dataset pages. This primarily checks that the max page count is displayed correctly and that selecting a page will load the correct items based on the backend.

### DIFF
--- a/frontend/packages/data-portal/e2e/pagination/testPagination.ts
+++ b/frontend/packages/data-portal/e2e/pagination/testPagination.ts
@@ -1,30 +1,65 @@
-import { ApolloClient, NormalizedCacheObject } from '@apollo/client'
+/* eslint-disable playwright/no-conditional-in-test */
+
 import { expect, Page, test } from '@playwright/test'
 import { getApolloClient } from 'e2e/apollo'
 import { E2E_CONFIG } from 'e2e/constants'
 import { TableValidator } from 'e2e/filters/types'
 import { goTo } from 'e2e/filters/utils'
+import { isNumber } from 'lodash-es'
 
 import { TestIds } from 'app/constants/testIds'
+
+import {
+  GetMaxPages,
+  PaginationFilter,
+  TestOptions as PaginationTestOptions,
+} from './types'
+import { getParamsFromFilter } from './utils'
 
 function getPagination(page: Page) {
   return page.getByTestId(TestIds.Pagination)
 }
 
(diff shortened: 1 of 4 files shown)


frontend/packages/data-portal/e2e/pagination/testPagination.ts
```diff
@@ -0,0 +1,103 @@
...
+      const node = getPagination(page)
+      const pageNodes = node.locator('li')
+
+      await pageNodes.nth(2).click()
+      await page.waitForSelector(rowLoadedSelector)
+
+      await validateTable({
+        client,
+        page,
+        pageNumber: 2,
```
kne42, line 78: would this test fail if the max pages were 1?

frontend/packages/data-portal/e2e/pagination/testPagination.ts
```diff
@@ -0,0 +1,103 @@
...
```
kne42, line 98: same as above, also what happens when there are no results?

kne42:
also we're missing tests on the arrows for the pagination e.g. if they disappear when at the first / last pages, if clicking on them gives us the correct page, etc.

frontend/packages/data-portal/e2e/pagination/testPagination.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    test('should show correct max page count', async ({ page }) => {
+      await goTo(page, url)
+
+      const node = getPagination(page)
+      const maxPages = await getMaxPages(client)
+      const pageNodes = node.locator('li')
+
+      const count = await pageNodes.count()
+      await expect(pageNodes.nth(count - 2)).toContainText(`${maxPages}`)
```
tihuan, line 46: hopefully this is not overengineering, but this block and the next test's test steps seem pretty similar, so an opportunity to DRY it up 😄 ?

tihuan, line 46: also just curious, what does the `count - 2` logic come from? Is it part of the SDS pagination spec or a business logic? 

Not sure if it's worth documenting that magic number 😄

tihuan (commented):
Looks v good! Love the extractions and reusable functions 👏 Just some small questions I have that are non blocking 😄 

Also I'll be out tomorrow till next Thursday! 

Have a great rest of your week!!

Issue #460, opened 2024-02-08
Update Annotation Method Type field


manasaV3:
Add the Method Type in the run page:
- [ ] Metadata panel ([design link here](https://www.figma.com/file/gwX2PjfGPfMrbEWMMJq6rW/H1-2024---Annotations-Table-Update?node-id=527%3A139631&mode=dev))
- [ ] Annotations Table ([design link here](https://www.figma.com/file/gwX2PjfGPfMrbEWMMJq6rW/H1-2024---Annotations-Table-Update?node-id=523%3A136924&mode=dev))

Blocked by #443 

Relates to #434 #432

manasaV3:
@codemonkey800  @kne42 Could you please point this.

manasaV3:
@junxini This is no longer blocked, and can be picked up by the frontend.

PR #797, opened 2024-06-08
chore: hook up method type to front-end

kne42:
#460


kne42:
https://dev-kira-1.cryoet.dev.si.czi.technology/runs/623

Issue #732, opened 2024-05-15
Object Count is not longer surfaced in the annotation metadata panel


manasaV3:
**Describe the bug**

While it is still a part of the [design](https://www.figma.com/design/gwX2PjfGPfMrbEWMMJq6rW/H1-2024---Annotations-Table-Update?node-id=527-139631&t=yKvvSt8sO62ul4hG-0), we are no longer surfacing the object count in the annotation metadata panel. 

**To Reproduce**
1. Go to https://cryoetdataportal.czscience.com/runs/426?metadata=annotation&tab=metadata

PR #804, opened 2024-06-13
fix: restore object count

codemonkey800:
#732

Adds the object count row that was missing from the annotation metadata drawer

## Demos

image


manasaV3 (approved):
🚢

PR #820, opened 2024-06-18
fix: Merge query filters properly

jgadling:
Using ```somedict.update()``` to merge filters was overwriting deeply nested filters instead of combining them. The `deepmerge` library can merge nested structures better.


manasaV3 (approved):

uermel (approved):
👍

PR #841, opened 2024-07-08
fix: update download button strings

### LINKED ISSUE #819: Single Run & single dataset download buttons have incorrect text
**Describe the bug**
On single run page, download button in the header says "Download Run"— should say "Download…"
On single dataset page, download button header says "Download Dataset" — should say "Download Dataset…"

**To Reproduce**
1. Go to any single run page & single dataset page
2. Look at download button in header

**Screenshots**

Image

 - OS: macOS 14.1
 - Browser : chrome Version 126.0.6478.62

**Additional context**
Team agreed that this button should have "Download…" so that users would know that there are different download options when they click the button. Saying "download run" is misleading - can't download the run -  and may prevent users from discovering download tomogram/all annotations

ehoops-cz:
Resolves #819 

The download buttons should indicate to the use that there is more information to be discovered.  This PR changes the text for single-run download to `Download...` and the text for single-dataset download to `Download Dataset...`

Screenshot 2024-07-08 at 12 03 28 PM
Screenshot 2024-07-08 at 12 03 38 PM


Janeece (approved):
LGTM, thanks @ehoops-cz !

Issue #816, opened 2024-06-18
Update copy in the method type tooltips


dgmccart:
Outdated copy is used in the method types tooltips. The updated copy is [found here](https://docs.google.com/spreadsheets/d/1QEPiDcjGJYWSZXsbRjwHd9aeZjbk_tvv2UQ4DKyMwxk/edit#gid=339544814&range=18:21) and below.

| Location | copy   | 
| :----: | :-------: | 
| Method Type | Method Type: Describes how the annotations were generated   | 
| Manual	| Manual: Annotations were created by hand |
| Automated | Automated: Annotations were generated using automated tools or algorithms without supervision |
| Hybrid	| Hybrid: Annotations were generated using a combination of automated and manual methods |

PR #847, opened 2024-07-08
feat: Update annotation method type tooltips text

bchu1:
https://github.com/chanzuckerberg/cryoet-data-portal/issues/816

image
image
image
image


bchu1:
> lgtm! rip we need to fix the tooltip centering but out of scope for this PR lol

Lol I didn't even notice that 😆

PR #851, opened 2024-07-09
ci: fix metadata drawer e2e

codemonkey800:
Follow up from #839 that fixes the metadata drawer E2E by passing the correct test data based on the ground truth status. 

https://github.com/chanzuckerberg/cryoet-data-portal/actions/runs/9862017314

### DIFF
--- a/frontend/packages/data-portal/e2e/metadataDrawer/utils.ts
+++ b/frontend/packages/data-portal/e2e/metadataDrawer/utils.ts
@@ -131,6 +131,10 @@ export function getAnnotationTestMetdata(
     objectState: annotation.object_state,
     objectDescription: annotation.object_description,
 
+    // Ground truth annotations show N/A for precision and recall because they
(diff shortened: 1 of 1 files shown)


ehoops-cz (approved):
It took me a second to realize why ground truth would not have precision and recall, so maybe a comment would be helpful.  Looks good though.

frontend/packages/data-portal/e2e/metadataDrawer/utils.ts
```diff
@@ -126,9 +130,13 @@ export function getAnnotationTestMetdata(
     objectShapeType: file?.shape_type,
     objectState: annotation.object_state,
     objectDescription: annotation.object_description,
+
+    // Ground truth annotations show N/A for precision and recall because they
+    // represent the correct or true labels for a dataset. Non ground truth
+    // annotations have these fields because they are compared against a ground
+    // truth annotation.
```
ehoops-cz, line 137: 🎉

PR #856, opened 2024-07-09
refactor: using page object model for e2e tests

ehoops-cz:
This PR tries to use the page object model for the metadataDrawer tests.  Note: some tests are failing - I will need to add Jeremy's ground truth changes into this change before merging.

Some pros/cons I found along the way:
Pro:
- putting helpers in the page object model forced me to come up with intelligible names for what was going on

Con:
- the tests are very similar to each other - I'm not sure I properly balanced being clear with being succinct.

What do we think of this approach?
Is it more readable? Is it harder to track things down?
Do we want less repetition? - if so, we could do this within the page object model or opt for sticking with functions.


frontend/packages/data-portal/e2e/metadataDrawer/metadata-drawer-page.ts
```diff
@@ -0,0 +1,389 @@
...
+export class MetadataDrawerPage extends PageObject {
+  // TODO: (ehoops) Which region does this belong in?
+  public getBoolString(value?: boolean): string {
+    return value ? 'True' : 'False'
+  }
+
+  // #region Navigate
+  // #endregion Navigate
+
+  // #region API
```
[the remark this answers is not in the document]
ehoops-cz: Honestly, I just copied them over from CZID and am not sure if it's something we should keep.  It looks like it adds sections to the right panel in VScode
Screenshot 2024-07-11 at 8 54 23 AM

frontend/packages/data-portal/e2e/metadataDrawer.test.ts
```diff
@@ -1,113 +1,131 @@
...
+import { test } from '@playwright/test'
+
+import { getApolloClient } from './apollo'
+import { E2E_CONFIG, SINGLE_DATASET_URL, SINGLE_RUN_URL } from './constants'
+import { MetadataDrawerPage } from './metadataDrawer/metadata-drawer-page'
+
+test.describe('Metadata Drawer', () => {
+  test.describe(`Single Dataset: ${SINGLE_DATASET_URL.replace(
+    E2E_CONFIG.url,
+    '',
```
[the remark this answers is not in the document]
ehoops-cz: I went with exporting just the path as well as the whole url.  lmk what you think

frontend/packages/eslint-config/e2e.cjs
```diff
@@ -5,6 +5,9 @@ module.exports = {
     // Useful for testing things sequentially
     'no-await-in-loop': 'off',
 
+    // Allow conditional statements in tests
+    'playwright/no-conditional-in-test': 'off',
```
ehoops-cz, line 9: Are we ok with this change?  It looks like we ignore this rule fairly regularly anyway...

frontend/packages/data-portal/e2e/metadataDrawer/metadata-drawer-page.ts
```diff
@@ -0,0 +1,389 @@
...
+  public getDatasetTestMetadata({
```
bchu1, line 32: Do we think these functions from utils.ts should stay in a separate file, since they're not technically tied to the "page" but rather just generic GQL queries/helpers?

frontend/packages/data-portal/e2e/page-objects/page-object.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  // NOTE: We should avoid using this method as much as possible. It's better to use
+  // the specific methods here: https://github.com/playwright-community/eslint-plugin-playwright/blob/main/docs/rules/no-wait-for-timeout.md
+  public async pause(seconds: number) {
+    // eslint-disable-next-line playwright/no-wait-for-timeout
+    await this.page.waitForTimeout(seconds * 1000)
+  }
+
+  public async scrollToElement(
```
bchu1, line 74: Maybe we can delete these `scroll` methods for now, I found them kind of unreadable/hard to debug in CZID, hopefully there are better options we can use in CryoET?

frontend/packages/data-portal/e2e/page-objects/metadataDrawer/metadata-drawer-page.ts
```diff
@@ -0,0 +1,171 @@
...
+      ) {
+        await this.expectMetadataTableCellToDisplayNotApplicable(label)
+        continue
+      }
+      await this.expectMetadataTableCellToDisplayEmpty(label)
+    }
+  }
+  // #endregion Validation
+
+  // #region Bool
```
bchu1, line 169: nit - delete

bchu1 (approved):

Issue #832, opened 2024-07-02
Refactor E2E tests to use page object model


codemonkey800:
The [page object model](https://playwright.dev/docs/pom) is an architecture pattern for playwright tests where we contain all of the interactions for a particular page within a class:

```ts
import { expect, type Locator, type Page } from '@playwright/test';

export class PlaywrightDevPage {
  readonly page: Page;
  readonly getStartedLink: Locator;
  readonly gettingStartedHeader: Locator;
  readonly pomLink: Locator;
  readonly tocList: Locator;

  constructor(page: Page) {
    this.page = page;
    this.getStartedLink = page.locator('a', { hasText: 'Get started' });
    this.gettingStartedHeader = page.locator('h1', { hasText: 'Installation' });
    this.pomLink = page.locator('li', {
      hasText: 'Guides',
    }).locator('a', {
      hasText: 'Page Object Model',
    });
    this.tocList = page.locator('article div.markdown ul > li > a');
  }

  async goto() {
    await this.page.goto('https://playwright.dev');
  }

  async getStarted() {
    await this.getStartedLink.first().click();
    await expect(this.gettingStartedHeader).toBeVisible();
  }

  async pageObjectModel() {
    await this.getStarted();
    await this.pomLink.click();
  }
}
```

Our E2E utils are primarily exported functions, making it a bit harder to maintain. The page object model would make it simpler to maintain and use.

ehoops-cz:
Spinning off new issues for the remaining work.  #962 and #963

PR #885, opened 2024-07-17
refactor: filter tests page object model

ehoops-cz:
Related to #832 

This PR refactors the filter tests to use the page object model.  The tests are a bit more involved than the `metadataDrawer`, so this PR also introduces an `actor` layer that aggregates steps from the `page object` layer.  The goal is to have each test contain a readable set of steps - enough that a new reader can easily understand what is being tested, but not so much that it's overwhelming.  There should be clear places to look for what's happening under the hood.


frontend/packages/data-portal/e2e/singleRunFilterOld.test.ts
ehoops-cz, line 1: Original test file.  This file calls helper functions rather than showing the test steps.

frontend/packages/data-portal/e2e/filters/testSingleSelectFilter.ts
ehoops-cz, line 1: This file contains a `testSingleSelectFilter` function, which is getting replaced as part of this PR.  It's not removed yet since this is a draft.

frontend/packages/data-portal/e2e/singleRunFilters.test.ts
tihuan, line 1: I like this direction this is going 💡 Drawing from what @codemonkey800 had before the page object refactor to keep the test DRY, perhaps we can still extract some common test steps as utility functions to reduce the verboseness of each test and make it easier to focus on the diff among different test cases?

E.g., the following two function calls seem to be paired very frequently, worth combining them in one util function?

```ts
const singleRunData = await filtersActor.getSingleRunDataWithParams({
        client,
        id: +E2E_CONFIG.runId,
        pageNumber: 1,
        url: SINGLE_RUN_URL,
        queryParam: QueryParams.ObjectName,
        value: E2E_CONFIG.objectName,
      })

      await filtersActor.expectAnnotationsTableToBeCorrect({ singleRunData })
```

tihuan (commented):
Page Object and Actor LGTM 💡 Also TIL about the `// region` comment thing to allow VSCode to collapse a code section. Pretty cool feature!

frontend/packages/data-portal/e2e/singleRunFilters.test.ts
```diff
@@ -0,0 +1,344 @@
...
+  // TODO: (ehoops) Add this actual test!
+  test('Annotation Author filter', async ({ page }) => {
+    const filtersPage = new FiltersPage(page)
+    await filtersPage.goTo('https://playwright.dev/')
+  })
+
+  test.describe('Object Name filter', () => {
+    test('should filter when selecting', async ({ page }) => {
+      const filtersPage = new FiltersPage(page)
+      const filtersActor = new FiltersActor(filtersPage)
```
bchu1, line 26: nit - wonder if we can have these 2 lines in the `beforeEach()` like with `client`

frontend/packages/data-portal/e2e/pageObjects/filters/filtersActor.ts
```diff
@@ -0,0 +1,170 @@
...
+
+import { FiltersPage } from './filtersPage'
+import {
+  getAnnotationRowCountFromData,
+  getExpectedFilterCount,
+  getExpectedTotalCount,
+  getExpectedUrlWithQueryParams,
+} from './utils'
+
+export class FiltersActor {
```
bchu1, line 16: Could you add some documentation on what is expected to be in an Actor and what isn't?

bchu1 (approved):

frontend/packages/data-portal/e2e/pageObjects/filters/filtersActor.ts
ehoops-cz: Done! Updated documentation is here: https://github.com/chanzuckerberg/cryoet-data-portal/wiki/Testing

frontend/packages/data-portal/e2e/pageObjects/filters/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+    ?.annotation_table.at(0)
+    ?.annotations.reduce((counter, annotation) => {
+      const objectShapeTypes = new Set()
+      for (const file of annotation.files) {
+        objectShapeTypes.add(file.shape_type)
+      }
+      counter[annotation.id] = objectShapeTypes.size
+      return counter
+    }, rowCounter)
+  return rowCounter
```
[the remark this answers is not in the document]
ehoops-cz: Doing it this was was to appease typescript to return `AnnotationRowCounter` rather than `AnnotationRowCounter | undefined`.  The downstream functions won't accept undefined.  I added the return type to the function definition.

PR #892, opened 2024-07-17
fix: Don't add annotation rows for shape types that are filtered out

### LINKED ISSUE #891: Object shape type filter returns displays annotations
**Describe the bug**
When I go to a single run page and filter by "InstanceSegmentation" (https://cryoetdataportal.czscience.com/runs/5994?object_shape=InstanceSegmentation) there is an annotation ID that has two rows, ID 8238. There is a line for "InstanceSegmentation" and a line for "SegmentationMask".  With the filter applied, only the "InstanceSegmentation" row should show.

**To Reproduce**
1. Go to ['A single run'](https://cryoetdataportal.czscience.com/runs/5994)
2. Click on the 'Object shape type' filter
3. Select "InstanceSegmentation" (also happens with "SegmentationMask")
4. See that the annotations table has the correct IDs, but there is an extra row if there are both an 'InstanceSegmentation" and "SegmentationMask" for the same ID.

**Screenshots**
Screenshot 2024-07-17 at 11 42 20 AM

 - Browser - any

**Additional context**
Looks like the filter is correctly getting the IDs that have the right segmentation, but then it returns everything for each of those IDs.

ehoops-cz:
Resolves #891 

In order to display a separate annotation row for each shape type, we expand the list of annotations on the frontend.  This adds a check to the expansion function for whether the shape type is filtered out.

filter-fix

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
@@ -14,7 +14,6 @@ import { MAX_PER_PAGE } from 'app/constants/pagination'
 import { AnnotationTableWidths } from 'app/constants/table'
 import { TestIds } from 'app/constants/testIds'
 import { useDownloadModalQueryParamState } from 'app/hooks/useDownloadModalQueryParamState'
-import { useFilter } from 'app/hooks/useFilter'
 import { useI18n } from 'app/hooks/useI18n'
 import { useIsLoading } from 'app/hooks/useIsLoading'
 import {
@@ -83,7 +82,6 @@ export function AnnotationTable() {
   const { toggleDrawer } = useMetadataDrawer()
   const { setActiveAnnotation } = useAnnotation()
   const { t } = useI18n()
-  const { annotation: annotationFilter } = useFilter()
 
   const { openAnnotationDownloadModal } = useDownloadModalQueryParamState()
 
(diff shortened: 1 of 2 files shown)


bchu1:
Hmm, I think the bug is actually in getRunById.server.ts#288, it's querying for all `annotations` that contain `files` with the `shape_type` but it's still selecting all `files`. The object shape filter should technically be on the `files` field not `annotations`

```
files(where: { object_shape: { _in: [...] } }) {
  format
  https_path
  s3_path
  shape_type
}
```

It'd prolly have to be a separate filter variable passed into `client.query()`

ehoops-cz:
> Hmm, I think the bug is actually in getRunById.server.ts#288, it's querying for all `annotations` that contain `files` with the `shape_type` but it's still selecting all `files`. The object shape filter should technically be on the `files` field not `annotations`
> 
> ```
> files(where: { object_shape: { _in: [...] } }) {
>   format
>   https_path
>   s3_path
>   shape_type
> }
> ```
> 
> It'd prolly have to be a separate filter variable passed into `client.query()`

I thought this, but the `run` object returned from gql has a single annotation in it with or without the filter.  In order to show a new row in the table, the annotation gets duplicated - this is what we want to display when there are no filters applied.  Should we change this behavior in the GQL layer?

ehoops-cz:
Ok, after a chat the new course of action is to add a fileFilter to `getRunById.server.ts`.  This will remove any files that do not pass the filters from the `run` response so we won't expand the annotation incorrectly.

bchu1 (approved):

PR #895, opened 2024-07-19
fix: Use match with substring for exception check in client tests

andy-sweet:
The `test_filter_on_object_raises_exceptions` test previously checked for an exception to have an exact test message. The schema must have changed a little since these were first written, so now these tests fail.

Instead of trying to match the whole string and updating it every time the schema changes, I used the `match` keyword to match a common substring that should be more stable.

I also ignored a small test file that's created as part of the test setup.


jgadling (approved):
Thanks for the fix!

PR #910, opened 2024-07-23
feat: Implement Python client code generator

### LINKED ISSUE #894: Python client models should be automatically derived from the GraphQL schema
The model classes in the Python client are manually maintained in [_models.py](https://github.com/chanzuckerberg/cryoet-data-portal/blob/6a38f11afe46213e247d204a272d6b49165e097b/client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py).

The class and field names are almost entirely defined in the GraphQL schema, which is [also in this repository](https://github.com/chanzuckerberg/cryoet-data-portal/blob/6a38f11afe46213e247d204a272d6b49165e097b/client/python/cryoet_data_portal/src/cryoet_data_portal/data/schema.graphql).

Maintaining this manually [has created documentation errors in the past](https://github.com/chanzuckerberg/cryoet-data-portal/pull/282), and may lead to bigger issues in the future when the schema changes.

The goal here is to automatically derive these model classes from the GraphQL schema.

andy-sweet:
Closes #894

Based on https://github.com/chanzuckerberg/cryoet-data-portal/pull/895

This adds a custom code generator for the Python client. The mapping between GraphQL types, model types, and fields is somewhat explicit, but doesn't require too much code. The custom download utilities are implemented in a separate stub module, since I could not think of a sensible way to generate that code.

After quickly trying some of the existing packages for generating Python GraphQL clients, my sense was that writing something specific for this client is probably easier than learning how to configure those tools for this particular client. This approach also has the advantage not imposing its own client and query style, which makes backwards compatibility easier to handle.

I ran the tests locally and they all passed first time, which probably means I ran them incorrectly or their coverage is lacking.

The code generator source (including the model stubs) is not that much smaller than the current manually maintained `_models.py`, but I think it may have caught some inconsistencies and missing descriptions in the schema.

Marking this as a draft, since this is still a work-in-progress, with its own lack of test coverage and documentation. At a minimum, we could use the inconsistencies found here to fix the manually maintained module.


andy-sweet:
I had to roll back some new Python syntax to support 3.7 here (e.g. `:=` walrus operator is from 3.8 onwards). I also couldn't easily setup a local Python 3.7 environment on my ARM mac since it's no longer available on the standard conda-forge channels.

We might want to consider bumping the min supported version to 3.8 at the same time as listing support and running tests against 3.11 and 3.12. Another option for testing is to just test the min and max versions, which tends to work well enough in practice.

client/python/cryoet_data_portal/src/cryoet_data_portal/_model_stubs.py
```diff
@@ -0,0 +1,191 @@
...
+from __future__ import annotations
+
+import os
+from typing import Iterable, Optional
+
+from cryoet_data_portal._file_tools import download_directory, download_https
+
+
+class Annotation:
+    files: Iterable[AnnotationFile]
```
andy-sweet, line 11: These are only needed for type checking the method bodies. We should probably compare the parsed fields to verify the typing of these is correct.

andy-sweet:
For comparison, see the output from running [datamodel-code-generator](https://koxudaxi.github.io/datamodel-code-generator/): https://gist.github.com/andy-sweet/7ae896ba079b9774ec8572a48208be57

After a little experimentation with that tool (and others), I ran the following command to generate that output.

```shell
datamodel-codegen --input data/schema.graphql --input-file-type graphql --output-model-type pydantic_v2.BaseModel --use-field-description --output _models.py
```

The output completely captures the whole schema including relationships, but is missing field descriptions and also seems to have some suspicious typing (e.g. `Tomograms.authors: List[tomogram_authors]`).

We'd likely have to do some wrapping/stubbing (as we do in this PR) to handle the model utility methods, so this might still be useful as intermediate output.

There may also be ways to configure the tool to do some of the work we need (e.g. templating, table -> model renames), but that's not clear to me yet and would require extra work.

client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
```diff
@@ -0,0 +1,231 @@
...
+    annotation_type: str
+    default_value: str
+
+
+def write_models(path: str) -> None:
+    schema = load_schema()
+    with open(path, "w") as f:
+        f.write(
+            dedent(
+                """\
```
andy-sweet, line 68: I should probably be using a templating tool (e.g. jinja2) instead of all these writes.

We could also consider defining a template per model, which could also be useful for handling the utility methods.

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -16,34 +20,34 @@
...
         organism_taxid (str): NCBI taxonomy identifier for the organism, e.g. 9606
         other_setup (str): Describe other setup not covered by sample preparation or grid preparation that may make this dataset unique in  the same publication
         related_database_entries (str): If a CryoET dataset is also deposited into another database, enter the database identifier here (e.g. EMPIAR-11445). Use a comma to separate multiple identifiers.
-        related_database_links (str): If a CryoET dataset is also deposited into another database, e.g. EMPIAR, enter the database identifier here (e.g.https://www.ebi.ac.uk/empiar/EMPIAR-12345/).  Use a comma to separate multiple links.
+        related_database_links (str): If a CryoET dataset is also deposited into another database, e.g. EMPAIR, enter the database identifier here (e.g.https://www.ebi.ac.uk/empiar/EMPIAR-12345/).  Use a comma to separate multiple links.
         release_date (date): Date when a dataset is made available on the Data Portal.
-        runs: List[Run] An array relationship with Run
+        runs (List[Run]): An array relationship
```
andy-sweet, line 50: This is the description of this relationship in schema.graphql and should probably be updated there (or whatever its ultimate source is).

This is a common theme for the list and item relationships.

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -106,91 +105,87 @@ def download_everything(self, dest_path: Optional[str] = None):
...
+        affiliation_identifier (str): A unique identifier assigned to the affiliated institution by The Research Organization Registry (ROR).
         affiliation_name (str): Name of the institutions an author is affiliated with. Comma separated
-        author_list_order(int): The order in which the author appears in the publication
-        corresponding_author_status (bool):Indicating whether an author is the corresponding author
-        dataset (Dataset): An object relationship with the dataset this author corresponds to
-        dataset_id (int): Numeric identifier for the dataset this author corresponds to
+        author_list_order (int): The order in which the author appears in the publication
+        corresponding_author_status (bool): Indicating whether an author is the corresponding author
+        dataset (Dataset): An object relationship
+        dataset_id (int): None
```
andy-sweet, line 117: This does not have a description in `schema.graphql`, so should be fixed there or its ultimate source.

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -253,58 +246,58 @@ def download_everything(self, dest_path: Optional[str] = None):
...
+        scale0_dimensions (str): comma separated x,y,z dimensions of the unscaled tomogram
+        scale1_dimensions (str): comma separated x,y,z dimensions of the scale1 tomogram
+        scale2_dimensions (str): comma separated x,y,z dimensions of the scale2 tomogram
         size_x (int): Number of pixels in the 3D data fast axis
         size_y (int): Number of pixels in the 3D data medium axis
-        size_z (int):  Number of pixels in the 3D data slow axis.  This is the image projection direction at zero stage tilt
+        size_z (int): Number of pixels in the 3D data slow axis.  This is the image projection direction at zero stage tilt
         tomogram_version (str): Version of tomogram using the same software and post-processing. Version of tomogram using the same software and post-processing. This will be presented as the latest version
-        tomogram_voxel_spacing (TomogramVoxelSpacing): An object relationship with a specific voxel spacing for this experiment run
-        type (str):  Tomogram purpose (ex: CANONICAL)
```
andy-sweet, line 288: This got lost in the generation process and should be fixed here.

We could consider defining a string enum to handle enum fields, though that's probably for another issue.

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -253,58 +246,58 @@ def download_everything(self, dest_path: Optional[str] = None):
...
-
-    id: int = IntField()
-    affine_transformation_matrix: str = StringField()
-    ctf_corrected: int = BooleanField()
+    ctf_corrected: bool = BooleanField()
     deposition_id: int = IntField()
     fiducial_alignment_status: str = StringField()
     https_mrc_scale0: str = StringField()
     https_omezarr_dir: str = StringField()
-    is_canonical: int = BooleanField()
```
andy-sweet, line 307: I assume this was a copy/paste error and that the attribute annotation should be `bool` (and not `int`) for `BooleanField`.

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -541,91 +526,91 @@ def download(self, dest_path: Optional[str] = None):
...
-
-    run: Run = ItemRelationship(Run, "run_id", "id")
+    _gql_type: str = "tiltseries"
 
     acceleration_voltage: int = IntField()
     aligned_tiltseries_binning: int = IntField()
     binning_from_frames: float = FloatField()
     camera_manufacturer: str = StringField()
     camera_model: str = StringField()
-    data_acquisition_software: float = FloatField()
```
andy-sweet, line 628: Looks like a copy/paste error. This works on `main` because the definiton of `FloatField` and `StringField` currently identical (i.e. they just extend `BaseField` without overriding or adding anything).

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
jgadling: Yep almost certainly a typo!

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
andy-sweet: Now fixed. The issue was that this an object type field, but not a model, which confused the current code generator logic. I included a workaround to specifically parse fields of model types, then handle everything else as a scalar, but that might be fragile.

andy-sweet:
I'm at a good stage to pause this PR. After some polishing, docs, and tests, I'd be fairly confident with the solution here for handling the short term needs of the Python client, and without breaking current usage.

But I need to check how the schema is evolving in the future (e.g. maybe there are more advanced relationships between fields?), and longer term plans to migrating the back-end (and if those would give us a Python client for free).

andy-sweet:
I also implemented this using jinja2 to replace the write formatting and model stubs (but keeping the schema parsing): https://github.com/chanzuckerberg/cryoet-data-portal/compare/main...andy-sweet:feat-python-client-codegen-jinja?expand=1

Overall, I think using jinja produces the more readable and understandable code, though we may want to do some type checking or static analysis of the generated code to ensure the fields and template methods are consistent.

andy-sweet:
When discussing this, we had the following questions and feedback.

Can we and should we automatically scrape the schema from the server rather than maintain the `schema.graphql` separately?
Yes. This should be doable with some existing GraphQL client packages. Ideally, we would have a two stage generation process: 1) Scrape the schema from the server; 2) Generate Python code from the schema.

How should we handle deprecated fields?
This is not handled right now. We should be able to manually mark certain fields and relationships as deprecated in Hasura. Then detect those as deprecated when processing the parsed schema in the code-gen module here. Then generate a property getter/setter that issues an appropriate deprecation warning.

For implementing the custom download methods, should we subclass the generated models?
This would make it easier to maintain those custom methods because typing information could be used while writing them. However, it has the downside of the making the generated code slightly harder to navigate, which for a public client API could be important.

jgadling:
I typically prefer using templating languages (and in Python it looks like Jinja is the norm) for this instead of imperatively writing files a few lines at a time:
- The templates are (mostly) declarative, so we can make changes to the WHAT of the output we want separately from the HOW we read GQL schemas and interpret them.
- There's more context in the templates (class definitions, imports, etc), so we can read them more like standard python code

andy-sweet:
> I typically prefer using templating languages

Agreed. I'll almost certainly just merge my approach using jinja back into this branch: https://github.com/chanzuckerberg/cryoet-data-portal/compare/main...andy-sweet:feat-python-client-codegen-jinja?expand=1

andy-sweet:
Addressed most of the feedback, except the subclass/template suggestion.

> **Can we and should we automatically scrape the schema from the server rather than maintain the `schema.graphql` separately?**
> Yes. This should be doable with some existing GraphQL client packages. Ideally, we would have a two stage generation process: 1) Scrape the schema from the server; 2) Generate Python code from the schema.

The most interesting update is that automatically scraping the schema from the server causes some of the tests to fail because of the field `primary_author_status`.

```
FAILED tests/test_get_client.py::test_relationships - gql.transport.exceptions.TransportQueryError: {'extensions': {'code': 'validation-failed', 'path': '$.selectionSet.annotation_authors.selectionSet.primary_author_status'}, 'message': "field 'primary_author_status' not found in type: 'annotation_authors'"}
FAILED tests/test_get_client.py::test_relationships_reverse - gql.transport.exceptions.TransportQueryError: {'extensions': {'code': 'validation-failed', 'path': '$.selectionSet.annotation_authors.selectionSet.primary_author_status'}, 'message': "field 'primary_author_status' not found in type: 'annotation_authors'"}
============================================================================================ 2 failed, 23 passed in 4.45s ============================================================================================
make: *** [test] Error 1
```

@jgadling or @manasaV3 : any idea what causes this inconsistency between what the server returns as a schema through introspection and what it can deal with as part of a request?

andy-sweet:
> @jgadling or @manasaV3 : any idea what causes this inconsistency between what the server returns as a schema through introspection and what it can deal with as part of a request?

My guess is because the client test-infra maintains its own schema definitions that were not kept in sync?

https://github.com/chanzuckerberg/cryoet-data-portal/tree/f21d83b6bc1fa9e77b46b4e4ae3fa9873069a730/client/python/cryoet_data_portal/tests/test_infra/hasura/migrations/cryoetdataportal

client/python/cryoet_data_portal/tests/test_infra/hasura/migrations/cryoetdataportal/1722547580676_alter_table_public_annotation_authors_add_column_primary_author_status/up.sql
```diff
@@ -0,0 +1,3 @@
+alter table "public"."annotation_authors" add column "primary_author_status" boolean
```
andy-sweet, line 1: I copied this from the corresponding change on the backend repo: https://github.com/chanzuckerberg/cryoet-data-portal-backend/commit/60fe5b436631fbcb594b4bcbe71d1fbee2782c0e

PR #911, opened 2024-07-23
feat: Refactor <TablePageLayout> to support multiple tabs

bchu1:
This is to support having both Annotations and Tomograms tabs on the Runs page.
 - Adds support for a header above the tabs.
 - Uses our existing `<Tabs>` component.
   - Only shows tabs if more than 1 is specified.
 - Supports a different pagination query param per tab.
   - This changes the query param on the Runs page from `page` to `annotations-page`. In a subsequent PR I'll add `tomograms-page`. 
   - It seems like we're currently pretty far from either annotations or tomograms getting to a point where they'd actually trigger pagination, but I'd prefer this to work properly from the get-go rather than find out there's a bug later.
 - Stores the currently open tab in URL params.
 - Make a few prop names clearer.
 - Allow `<Tabs>` to take in components as labels and `number`s (i.e. indexes) as values.
 - Also enable hoisting in ESLint (which TS checks anyways).

### DIFF
--- a/frontend/packages/data-portal/app/components/TablePageLayout.tsx
+++ b/frontend/packages/data-portal/app/components/TablePageLayout.tsx
@@ -97,28 +97,28 @@ function TablePageTabContent({
   filteredCount,
   table,
   noResults,
-  pageQueryParamKey: pageQueryParam = QueryParams.Page,
+  pageQueryParamKey = QueryParams.Page,
   totalCount,
   countLabel,
 }: TableLayoutTab) {
   const [searchParams, setSearchParams] = useSearchParams()
-  const pageQueryParamValue = +(searchParams.get(pageQueryParam) ?? '1')
+  const pageQueryParamValue = +(searchParams.get(pageQueryParamKey) ?? '1')
 
   useEffect(() => {
     if (Math.ceil(filteredCount / MAX_PER_PAGE) < pageQueryParamValue) {
       setSearchParams(
         (prev) => {
-          prev.delete(pageQueryParam)
+          prev.delete(pageQueryParamKey)
           return prev
         },
         { replace: true },
       )
     }
-  }, [filteredCount, pageQueryParamValue, pageQueryParam, setSearchParams])
+  }, [filteredCount, pageQueryParamKey, pageQueryParamValue, setSearchParams])
 
   function setPage(nextPage: number) {
     setSearchParams((prev) => {
-      prev.set(pageQueryParam, `${nextPage}`)
+      prev.set(pageQueryParamKey, `${nextPage}`)
       return prev
     })
   }
--- a/frontend/packages/data-portal/app/components/TablePageLayout.tsx
+++ b/frontend/packages/data-portal/app/components/TablePageLayout.tsx
@@ -1,6 +1,7 @@
 import { Pagination } from '@czi-sds/components'
 import { useSearchParams } from '@remix-run/react'
-import { ReactNode, useEffect, useMemo, useState } from 'react'
+import { toNumber } from 'lodash-es'
+import { ReactNode, useEffect, useMemo } from 'react'
 
 import { TABLE_PAGE_LAYOUT_LOG_ID } from 'app/constants/error'
 import { MAX_PER_PAGE } from 'app/constants/pagination'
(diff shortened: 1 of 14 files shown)


frontend/packages/data-portal/app/components/TablePageLayout.tsx
```diff
@@ -1,57 +1,118 @@
...
-  table: ReactNode
-  title: string
-  totalCount: number
-  type: string
-}) {
+  countLabel,
+}: TableLayoutTab) {
   const [searchParams, setSearchParams] = useSearchParams()
-  const page = +(searchParams.get('page') ?? '1')
+  const page = +(searchParams.get(pageQueryParam) ?? '1')
```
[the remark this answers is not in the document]
bchu1: This prolly won't apply for some time, but if the user switches tabs and then performs an action that causes another load, whatever the page was that they were on in the previous tab will get reset to 1, which might be jarring if they're switching back and forth between tabs

frontend/packages/data-portal/app/components/TablePageLayout.tsx
```diff
@@ -1,57 +1,118 @@
...
+
+/** Standard page structure for browsing + filtering list(s) of objects. */
 export function TablePageLayout({
+  header,
+  tabs,
+  tabsTitle,
   downloadModal,
   drawers,
+}: TablePageLayoutProps) {
+  const [activeTabIndex, setActiveTabIndex] = useState<number>(0)
```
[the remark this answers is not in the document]
bchu1: Done.

ehoops-cz:
Do we think people may have the `page` query param bookmarked?  If so, do we want a redirect to `annotations-page`?  This may be overkill, but I think it's worth considering.

frontend/packages/data-portal/app/components/TablePageLayout.tsx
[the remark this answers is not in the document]
ehoops-cz: Is `page` currently for paginating the table?

frontend/packages/data-portal/app/components/TablePageLayout.tsx
```diff
@@ -4,54 +4,121 @@ import { ReactNode, useEffect, useMemo } from 'react'
...
         (prev) => {
-          prev.delete('page')
+          prev.delete(pageQueryParam)
           return prev
         },
         { replace: true },
       )
     }
-  }, [filteredCount, page, setSearchParams])
+  }, [filteredCount, page, pageQueryParam, setSearchParams])
```
ehoops-cz, line 117: It's confusing that `page` and `pageQueryParam` are two different things.  Can we rename one?  
`tableTypeQueryParam`? `dataTypeQueryParam`?  🤔

bchu1:
> Do we think people may have the `page` query param bookmarked? If so, do we want a redirect to `annotations-page`? This may be overkill, but I think it's worth considering.

Definitely a good thing to keep in mind. Right now nothing on the runs page ever exceeds 1 page so we should be good.

frontend/packages/data-portal/app/components/TablePageLayout.tsx
bchu1: Renamed both.

frontend/packages/data-portal/app/routes/browse-data.datasets.tsx
```diff
@@ -45,19 +45,23 @@ export default function BrowseDatasetsPage() {
...
+          noResults: (
+            <NoResults
+              title={i18n.filterNoResultsFound}
+              description={i18n.filterTooRestrictive}
+              actions={<Button onClick={reset}>{i18n.clearFilters}</Button>}
+            />
+          ),
+          filteredCount: filteredDatasetCount,
+          totalCount: datasetCount,
+          countLabel: i18n.datasets,
```
kne42, line 62: while you're here, would you mind updating to using `useI18n()`/`t`?

kne42 (approved):

frontend/packages/data-portal/app/routes/browse-data.datasets.tsx
bchu1: Done.

PR #914, opened 2024-07-24
fix: mdx page dates

codemonkey800:
Fixes issue with dates on MDX pages not reflecting the current file's last modified date. The root cause was that we weren't passing in the path when listing the most recent commit, resulting in reading the most recent commit for the entire repo.

### Before

image

### After

image


bchu1 (approved):

kne42 (approved):
i have some concerns leaving this to automation in the first place (conversation held offline) but will approve this as the fix to the current problem

Issue #887, opened 2024-07-17
Improve error message when downloading to a new folder location


kandarpksk:
When dest_path in any of the download methods in the API is a new folder location, i.e. a folder that doesn't currently exist, the error message shown is confusing: “Recursive downloads require a base directory”

Bento007:
How do we feel about creating the base path if it doesn't already exist. That makes one less step for our users to take, or if we don't want to assume what the users want, we can add a flag to make auto creating the directory optional.

Bento007:
> How do we feel about creating the base path if it doesn't already exist. That makes one less step for our users to take, or if we don't want to assume what the users want, we can add a flag to make auto creating the directory optional.

I see that we do [create the dest_path](https://github.com/chanzuckerberg/cryoet-data-portal/blob/3cf96cb6b785b1d8e72baef57f42b7b6706bca69/client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py#L90-L90) if `recursive_from_prefix` is False. What was the reasoning behind this?

manasaV3:
> I see that we do [create the dest_path](https://github.com/chanzuckerberg/cryoet-data-portal/blob/3cf96cb6b785b1d8e72baef57f42b7b6706bca69/client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py#L90-L90) if recursive_from_prefix is False. What was the reasoning behind this?

I believe this is because, if it is not a recursive download we are downloading a complete directory or else it is pulling down a single file which should be easier to identify.  So, creating the directory helps consolidate all of the contents what we are downloading for a recursive download.

kandarpksk:
My original understanding was that the error is thrown whenever the destination path doesn't exist. But from Manasa's comment, it seems like that's not always true. I think we should just make it consistent and create the missing folder in the path.

manasaV3:
@kandarpksk to confirm, we won't be creating a directory in all the cases when the user doesn't provide a destination path. (ie) we will only create a directory for the cases where we are downloading a directory, like in the case of a zarr file or if we are downloading an entire run. 
If the user explicitly provides a destination path that is not a directory, I think we shouldn't be creating that directory, but should provide a user understandable error that explains the path they provided is not a directory, so we can't download files into that path. Let me know what you think.

kandarpksk:
Makes sense. I certainly don't think we should always create a directory.

PR #940, opened 2024-07-25
fix: create recursive_from_prefix path if it does not exist

Bento007:
# Reason
- #887 

# Changes
- create the recursive_from_prefix directory if it does not exist.
- add docstring to `get_destination_path`
- add unit tests for `get_destination_path`


client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
```diff
@@ -57,12 +57,29 @@ def get_destination_path(
...
+
+    """
     if not dest_path:
         dest_path = os.getcwd()
     dest_path = os.path.abspath(dest_path)
 
     if not os.path.isdir(dest_path) and recursive_from_prefix:
-        raise ValueError("Recursive downloads require a base directory")
+        raise ValueError(
+            f"The destination path '{dest_path}' does not exist. Recursive downloads require an existing base directory.",
```
Bento007, line 81: @kandarpksk what are your thoughts on the modified error message.

manasaV3 (approved):
LGTM

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
kandarpksk: I just read recent comments in the [issue](https://github.com/chanzuckerberg/cryoet-data-portal/issues/887), and think we should just create the missing folder in all cases.

If we don't go with that, I find the updated message still confusing. I think the language we use should be from the perspective of the API user, who is either downloading a file (with say, download_mrcfile) or a folder (with say, download_everything). In which case is the download recursive? (I'd ordinarily assume that the folder download is recursive, but from [Manasa's comment](https://github.com/chanzuckerberg/cryoet-data-portal/issues/887#issuecomment-2251089496), it seems to me that in that case, this error is not thrown!)

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
```diff
@@ -54,15 +54,27 @@ def download_https(
...
+
+    """
     if not dest_path:
         dest_path = os.getcwd()
     dest_path = os.path.abspath(dest_path)
 
-    if not os.path.isdir(dest_path) and recursive_from_prefix:
-        raise ValueError("Recursive downloads require a base directory")
+    if not os.path.isdir(dest_path):
+        os.makedirs(dest_path)
```
manasaV3, line 77: When the dest_path is None, it should point to the working directory, and will pass this check. So, the only time this exception would show up would be when the user explicitly provides an invalid value and we should error on it. I think the error message could be more verbose in explaining the issue.  cc: @kandarpksk 

Also, [this line](https://github.com/chanzuckerberg/cryoet-data-portal/pull/940/files#diff-63d5c8af96144c2f98bba0f8110bc718612219f91e9b97e9b40e3d74a9a9bf73R85) should already be creating the recursive directory structure.

manasaV3 (changes requested):

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
```diff
@@ -54,15 +54,27 @@ def download_https(
 
 def get_destination_path(
     url: str,
-    dest_path: Optional[str],
+    dest_path: Optional[str] = None,
```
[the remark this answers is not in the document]
Bento007: > Is there a case, where we are calling this method without a value for dest_path?

It's only used in [one](https://github.com/chanzuckerberg/cryoet-data-portal/blob/e9046e43cb7580daaefa7eb329b82aecb208f805/client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py#L118-L118) place, and dest_path is provided.

I see, you're using Optional to mean String or None while at the same time enforcing that dest_path must be a parameter.

client/python/cryoet_data_portal/src/cryoet_data_portal/_file_tools.py
Bento007: So error when an invalid name is provided? Invalid meaning illegal path characters or maybe a path we can't write to?

manasaV3 (approved):

Issue #433, opened 2024-02-01
Annotation table section dividers


junxini:
**Description**: Table section dividers are added to more easily identify which annotations are ground truth

**Tasks:**

- [x] Add 2 section dividers with count: "# Ground Truth Annotations" and "# Other Annotations"
  - "#" represents the total count of annotations in that section (not just what's on the page)
- [x] Section overflows onto the next page (when using pagination), count in dividers should remain the same
- [x] Applying filters will:
  - [x] update the count within the section dividers to reflect
  - [x] ~~hide the section divider if there are no results in that section~~ no longer the case, see below

**Link to design**: https://www.figma.com/design/Kl03vJ6BID3ztPb8B0fsel/Annotation-Table-Updates---Dividers?node-id=1-6108&m=dev

manasaV3:
Here is a sample query for getting some of the basic information for this page:

```
query annotation_table_query($run_id: Int) {
  annotations(order_by: {ground_truth_used: desc, deposition_date: desc}, where: {tomogram_voxel_spacing: {run_id: {_eq: $run_id}}}) {
    id
    ground_truth_status
    deposition_date
    files_aggregate(distinct_on: shape_type) {
      nodes {
        shape_type
      }
    }
    confidence_precision
    confidence_recall
    object_name
    authors(order_by: {author_list_order: asc}) {
      author_list_order
      name
      id
      orcid
    }
  }
  ground_truth_count:annotations_aggregate(where: {tomogram_voxel_spacing: {run_id: {_eq: $run_id}}, ground_truth_status: {_eq: true}}) {    
    aggregate {
      count(columns: id)
    }
  }
  other_annotation_count:annotations_aggregate(where: {tomogram_voxel_spacing: {run_id: {_eq: $run_id}}, ground_truth_status: {_eq: false}}) {    
    aggregate {
      count(columns: id)
    }
  }
}

```

junxini:
> currently blocked, we need backend support to query the ground truth and non-ground truth annotation counts cc: @manasaV3

@manasaV3 Jeremy mentioned there's a backend ticket for this, could you help link when you have a chance?

manasaV3:
[...]

PR #949, opened 2024-07-26
feat: Add ground truth dividers to Annotations table and correct all annotation counts

bchu1:
https://github.com/chanzuckerberg/cryoet-data-portal/issues/433

Demo: https://dev-annotation-dividers.cryoet.dev.si.czi.technology/runs/241

@codemonkey800's original PR: https://github.com/chanzuckerberg/cryoet-data-portal/commit/3e17d8920716028d5403d160b92aa0620659b74d

 - Displays a divider before each ground truth/non ground truth sequence of rows in the table.
   - Also displays dividers with 0 count if that section is empty.
   - See the issue for discussions re: various edge cases.
 - Changes all counts (total, filtered, ground truth, and non ground truth) to count the unique `shape_type` and `annotation_id` pairs represented in `annotation_files` using `distinct_on` to match the number of rows displayed in the tables.
   - Removes the need to sum counts on the frontend.

I'm not super satisfied with having to have 2 callbacks in AnnotationsTable.tsx for displaying dividers, but couldn't really think of something better that still worked well with react-table and satisfied the UX requirements for displaying dividers when there are no rows for that section. Making the data type a union between `Annotation` and something else that represents a divider messes up the types of the column helper, and allowing `<Table>` to take a wrapper component would be the same problem but probably even uglier because it would have to take 2, one to wrap `<table>` and one to wrap the `<tr>`s. If I had to I'd probably pick the union type option and just send down the actual divider `ReactNode`s in the `data` array as rows, but that just feels like abusing react-table...


Janeece (approved):
@bchu1 Looks and works great, thank you!

Issue #522, opened 2024-02-29
Add a custom user-agent to python api client


manasaV3:
To help identify requests sent by the python api client is access logs, we should add a customer user-agent.

bnelson-czi:
We currently can't differentiate API vs other`boto3` usage in our s3 logs, so it would be useful to enable this for improved analytics.

PR #966, opened 2024-07-29
feat: add user agent to client requests

Bento007:
# Reason
- To make identifying the user agent who made a request to s3 or the portal backend easier to identify.
- #522 

# Changes
- Move __version__ to _version.py so it can be used in USER_AGENT. 
- Add _constants.py to store the USER_AGENT use in boto3 and http requests
- Add python version, and cryoet-data-portal client version to USER_AGENT.


manasaV3:
@bnelson-czi Wanted to put this in your radar. The user agent might soon be changing for the client.

manasaV3 (approved):

PR #968, opened 2024-07-30
feat: Update cell strain links.

### LINKED ISSUE #912: Update cell strain/cell line link
Currently this field maps out to:

```
"https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id={Dataset.cell_strain_id}"
```

with the cleaning and standardizing these values in the backend, this link does not work for all cases. It needs to be updated to: 

If `Dataset.cell_strain_id` matches the regex pattern `WBStrain[0-9]{8}`, we want to link to:

```
"https://wormbase.org/species/c_elegans/strain/{Dataset.cell_strain_id}"
```

For example, WBStrain00050783 should link to https://wormbase.org/species/c_elegans/strain/WBStrain00050783.

otherwise, we can link to:

```
"http://purl.obolibrary.org/obo/{Dataset.cell_strain_id.replace(":", "_")}"
```

ehoops-cz:
Resolves #912 

Rather than adding additional logic to the InfoLink component that would only be used for cell strains, this adds a specific CellStrainInfoLink.  This PR also moves the InfoLink components to their own files.  

Lastly, this PR uses the fractal file structure - it fits here anyway, but it would be great to know if there's broader buy in for using this pattern.  If so, I'll move over the Hygen template from CZID so we won't have to manually set up the folder/file structure for each component we add.


ehoops-cz:
@daniel-ji when are the backend changes going out for this?  The links will be broken for a while if this goes out too early, so maybe we can coordinate.

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/components/CellStrainInfoLink/index.ts
```diff
@@ -0,0 +1 @@
+export { CellStrainInfoLink } from './CellStrainInfoLink'
```
ehoops-cz, line 1: The index files are super small, but they make the import paths look nicer

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/components/CellStrainInfoLink/CellStrainInfoLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+  value?: string | null
+  id?: string | null
+}) {
+  if (!value) {
+    return <span>--</span>
+  }
+
+  if (id) {
+    const isWormbase = WORMBASE_PATTERN.test(id)
+    const isOBO = !isWormbase
```
ehoops-cz, line 19: I did this so the logic below will match the logic in InfoLink.tsx.  It felt weird to have OBO be the positive case in one and the negative case in the other.  Open to feedback here.

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/components/CellStrainInfoLink/CellStrainInfoLink.tsx
bchu1, line 1: Is it possible to just have 1 link component that handles `OBO`, `WORMBASE`, and `NIH`?

bchu1 (approved):

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/components/CellStrainInfoLink/CellStrainInfoLink.tsx
daniel-ji: To confirm, values that use NIH should be only numbers ([0-9]+), WORMBASE is /WBStrain[0-9]{8}/, and OBO should be anything else ([a-zA-Z]+:[0-9]+). I guess if it doesn't match any, maybe we shouldn't link?

Maybe to improve this actually, we can also make any value that matches NCBITaxon:[0-9]+ to be NIH linked. But we would have to pop off "NCBITaxon:" and only keep the numeric part to be used for the link 'https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=' (This is a minor improvement for the sake of consistency, not a blocker)

daniel-ji:
Maybe a better way to approach this:

in datasetInfoLinks.ts:

```javascript
NIH_ONTOLOGY_PATTERN = /NCBITaxon:[0-9]+/ 
OBO_PATTERN = /[a-zA-Z]+:[0-9]+/
```

in InfoLink.tsx or something like that (no need for a separate component?)
```javascript
// NOTE: no need to pass in isOBO parameter, we just do the checking in this component
if (id) {
	let link = undefined;
	if !isNaN(id) {
		link = `${NIH}${id}`;
	} else if id.match(NIH_ONTOLOGY_PATTERN) {
		link = `${NIH}${id.replace("NCBITaxon:","");
	} else if id.match(WORMBASE_PATTERN) {
		link = `${WORMBASE}${id.replaceAll(':', '_')}
	} else if id.match(OBO_PATTERN) {
		link = `${OBO}${id.replaceAll(':', '_')}`;
	} 
	// don't link otherwise
	...
}
```
Thoughts?

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/components/CellStrainInfoLink/CellStrainInfoLink.tsx
ehoops-cz: It is, but the default for these links is OBO and the default for the rest of them is NIH, so it would end up with some weird logic using isOBO or isStrain I guess

ehoops-cz: ok! after chatting with Daniel, this seems possible and we can use the string itself to differentiate where the links should go to.

daniel-ji:
Also one small thing, but it might be better to refer to everything that's NIH right now to be NCBI? Since NCBI is the specific organization that manages the taxonomy IDs, while NIH is the parent organization and they have various other suborgs that have different IDs / links. So instead of `NIH` as a constant, maybe `NCBI` and like `NCBI_PATTERN` instead of `NIH_PATTERN` or something like that. Thanks!

PR #970, opened 2024-07-30
fix: avoid double-encoding spaces in breadcrumb url

### LINKED ISSUE #814: Filter carry-back fails when using breadcrumb navigation if filter contains a space
**Describe the bug**
As in title. Using "All Datasets" breadcrumbs leads to wrong filter being applied when back on the "All Datasets" page. See video below. 

**To Reproduce**
1. Go to [browse datasets](https://cryoetdataportal.czscience.com/browse-data/datasets)
2. Apply filter "Homo sapiens"
3. Click on any dataset
4. Click on "All Datasets" breadcrumb

Filter now contains escape character `%20`.

**Screenshots**

https://github.com/chanzuckerberg/cryoet-data-portal/assets/6641113/bbf0f2f2-6ef5-4d79-8542-99ff289feace

 - OS: Sonoma 14.5
 - Browser chrome

ehoops-cz:
Resolves #814 

~This manually unencodes the space character and replaces `%20` with `+`.  It uses `encodeURIComponent`, but no longer uses the `URLSearchParams` object.  This was necessary to avoid re-encoding the `+` character.  It works, but the approach seems a bit hacky.~

Now just removing the encoding and it works.

Jul-30-2024 11-16-16

### DIFF
--- a/frontend/packages/data-portal/app/components/Breadcrumbs.tsx
+++ b/frontend/packages/data-portal/app/components/Breadcrumbs.tsx
@@ -12,9 +12,7 @@ import { cns } from 'app/utils/cns'
 
 function encodeParams(params: [string, string | null][]): string {
   const searchParams = new URLSearchParams(
-    (params.filter((kv) => isString(kv[1])) as string[][]).map((kv) =>
-      kv.map(encodeURIComponent),
-    ),
+    params.filter((kv) => isString(kv[1])) as string[][],
   )
 
   return searchParams.toString()


bchu1:
Could you link the bug in the description?

frontend/packages/data-portal/app/components/Breadcrumbs.tsx
```diff
@@ -11,13 +10,23 @@ import {
 import { cns } from 'app/utils/cns'
 
 function encodeParams(params: [string, string | null][]): string {
-  const searchParams = new URLSearchParams(
-    (params.filter((kv) => isString(kv[1])) as string[][]).map((kv) =>
-      kv.map(encodeURIComponent),
```
bchu1, line 16: Wonder what would happen if we just removed this `encodeURIComponent()`. That might fix it?

ehoops-cz: It does 👍

ehoops-cz: I'm a bit hesitant that we will have weird characters that we want to escape though.  Thoughts @codemonkey800 ?

bchu1: I think `URLSearchParams` should handle the escaping automatically

ehoops-cz: I thought so too, but I'm confused about how this works if it's escaped

bchu1: I think `useSearchParams()` unescapes stuff. Right now the code is double escaping ("Homo sapiens" turns into `organism=Homo%2520sapiens`)

ehoops-cz: Ya, I also saw the double escape.  Sounds good - I'm going to go ahead and merge.

bchu1 (approved):

Issue #942, opened 2024-07-25
Flakey unittests for client


Bento007:
**Describe the bug**
GHA job test-client-python is flakey.
### Error
```
>           raise MaxRetryError(_pool, url, error or ResponseError(cause))
E           urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=8080): Max retries exceeded with url: /v1/graphql (Caused by ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')))
```
**To Reproduce**
open a PR and wait for *test-client-python* to run.

**Screenshots**
![Screenshot 2024-07-25 at 12 07 53 PM](https://github.com/user-attachments/assets/13bbe62c-8dde-4931-b180-29fdd58ef7f4)

andy-sweet:
I'm running into this more frequently in https://github.com/chanzuckerberg/cryoet-data-portal/pull/1008 where I'm adding a few more tests that hit the fake endpoint used for testing.

@Bento007 : not sure if you already start working on this, but as these tests are some of the first tests to run under `pytest`, I wondered if adding `--wait` to `docker compose up -d` might be enough to fix the flakiness here?

If I understood the docker compose docs correct that should wait until all containers are healthy/running, otherwise the only guarantee is that have started.

andy-sweet:
Unsure how long the logs will last, but here's an example of a failed run on my branch: https://github.com/chanzuckerberg/cryoet-data-portal/actions/runs/10427079368/job/28881156801?pr=1008#step:5:1134

andy-sweet:
Since it would be good to fix this before merging https://github.com/chanzuckerberg/cryoet-data-portal/pull/1008 , I'm going to make a PR to try to fix this.

PR #1008, opened 2024-08-08
feat: Generate Python client code using GraphQL introspection

### LINKED ISSUE #894: Python client models should be automatically derived from the GraphQL schema

andy-sweet:
Closes #894

This adds a custom code generator for the Python client. The mapping between GraphQL types, model types, and fields is somewhat explicit, but doesn't require too much code. The code generation updates the schema and then parses it using de-factor GraphQL python library (which we already depend on). We then use Jinja2 templates to populate the models module.

I [tried some existing packages for generating Python GraphQL clients](https://github.com/chanzuckerberg/cryoet-data-portal/issues/894#issuecomment-2239561088), and my sense was that writing something specific for this client is probably easier than learning how to configure those tools for this particular client. This approach also has the advantage not imposing its own client and query style, which makes this compatible with the existing API by design.

I have done a little manual testing, but testing is mostly through the newly modified or added Python tests. These include checking that missing item relationships are handled gracefully by the client, which is at least a temporary requirement as the ingestion runs and may be a more permanent one.

As the prod GraphQL server has been updated ahead of the Python client, these changes also include the corresponding changes to the models module and GraphQL schema file. As well as the associated test infrastructure changes.


client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
```diff
@@ -0,0 +1,280 @@
...
+    "Boolean": ("BooleanField()", "bool"),
+    "Float": ("FloatField()", "float"),
+    "Int": ("IntField()", "int"),
+    "String": ("StringField()", "str"),
+    "date": ("DateField()", "date"),
+    "numeric": ("FloatField()", "float"),
+    "_numeric": ("StringField()", "str"),
+    "tomogram_type_enum": ("StringField()", "str"),
+    # TODO: check if I should make a JsonField to handle this.
+    "json": ("StringField()", "str"),
```
andy-sweet, line 39: Unsure how this value should be parsed into `Annotation.method_links`.

Should this be a `str` (as now) or a `dict`? I still need to look at real live values to understand this.

client/python/cryoet_data_portal/tests/test_codegen.py
```diff
@@ -0,0 +1,24 @@
+from graphql import GraphQLObjectType, GraphQLSchema
+
+from cryoet_data_portal._codegen import SCHEMA_PATH, load_schema, parse_fields
+
+
+def test_load_schema():
```
andy-sweet, line 6: I still need much more test coverage for code generator. I think just a few more tests for each of the public functions might be good enough.

client/python/cryoet_data_portal/tests/test_get_client.py
```diff
@@ -6,6 +6,12 @@ def test_relationships(client) -> None:
     ds = datasets.pop()
     assert ds
     assert ds.authors.pop()
+    depo = ds.deposition
```
andy-sweet, line 9: I modified these tests to account for depositions and deposition authors in a way that seems consistent with what came before.

I think these could generally be more meaningful, but this may not be the right time to think too deeply. Open to ideas though.

andy-sweet:
A few things are blocking this from be reviewable and thus mergeable.

1. I should increase code coverage by covering the code generator. I think we should be able to reuse the existing test infrastructure to do this in a fairly meaningful way without needing to make new complex mocks or stubs.

2. It seems like the `deposition_id` can be `None` in production. Assuming this is not temporary, I need to add some logic to the relationship field types to handle that case. We should also have at least one test to cover this too.

3. I need to determine what the type and value of `Annotation.method_links` should be (i.e. `str` vs. `dict`) and if that needs any special handling for the `dict`.

Ideally this PR would be much smaller, but I'm not sure how easy that is given that this mostly generates code/data and quite a few things have changed on prod.

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -704,15 +712,104 @@ def download_mrcfile(
...
+        "Annotation",
+        "id",
+        "deposition_id",
+    )
+    authors: List[DepositionAuthor] = ListRelationship(
+        "DepositionAuthor",
+        "id",
+        "deposition_id",
+    )
+    dataset: List[Dataset] = ListRelationship("Dataset", "id", "deposition_id")
```
andy-sweet, line 751: Note that `dataset` is the name in the schema and not `datasets` as we might hope here. I assume it's that to avoid a name clash, but I'm not sure. Not ideal, but we can probably live with it. We could also provide a `datasets` property alias in a Jinja template to improve the situation.

client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
manasaV3: This would be a `dict`, as it would be a JsonField. :)

andy-sweet: I removed it for now (so it's ignored) since @jgadling was unsure if this should be live and public. My plan is to add it in a follow-up PR since this one is more than large enough!

andy-sweet:
This is ready for review now.

The PR is *much* larger than I'd like (though almost all the delta is auto-generated in `_models.py`, `schema.graphql`, and `up.sql`), but I am unsure how to split this up without setting up new fakes/stubs that may take a significant amount of time and probably are not worth it in the long term.

There are two practical issues blocking this from being merged.

1) The tests I added exacerbate the flakiness described in #942, so we should almost certainly fix that before merging this.
2) One of my fancy tests that programmatically imports a temp module seems to break codecov. I should either change the test or find the right way to make codecov ignore that module.

client/python/cryoet_data_portal/Makefile
```diff
@@ -20,3 +20,11 @@ test:
...
 		export BOTO_SIGNATURE_VERSION=s3v4; \
 		pytest -vvv -s . $(TEST)
+
+.PHONY: codegen
+codegen:
+	python -m cryoet_data_portal._codegen
+	# Need to run pre-commit twice because black and ruff fight with each other.
+	# Ignore the return code because that is non-zero when pre-commit applies a fix.
+	-pre-commit run --files src/cryoet_data_portal/_models.py src/cryoet_data_portal/data/schema.graphql
+	-pre-commit run --files src/cryoet_data_portal/_models.py src/cryoet_data_portal/data/schema.graphql
```
jgadling, line 30: Should we just use ruff? afaik newer versions of ruff make black kindof unnecessary?

client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
```diff
@@ -63,10 +73,18 @@ class Dataset(Model):
     cell_type_id: str = StringField()
     dataset_citations: str = StringField()
     dataset_publications: str = StringField()
+    deposition: Deposition = ItemRelationship("Deposition", "deposition_id", "id")
```
jgadling, line 76: Switching to strings instead of passing classes around is nice! I forgot that the Relationship class can handle this!

client/python/cryoet_data_portal/Makefile
andy-sweet: I think @kne42 originally added these about a year ago, so checking if we can consolidate to just ruff since it may have improved rule coverage since then.

jgadling: I'm in favor of just using one tool unless there's something major we'd be missing out on. @manasaV3 ?

jgadling (approved):
This looks great! I think we'll need to update the docs templates to add depositions / deposition authors when we get a chance but that can be handled in a later PR.

andy-sweet:
> This looks great! I think we'll need to update the docs templates to add depositions / deposition authors when we get a chance but that can be handled in a later PR.

If there's no existing issue I'll add an one after merging this to make sure that doesn't get lost.

andy-sweet:
> One of my fancy tests that programmatically imports a temp module seems to break codecov. I should either change the test or find the right way to make codecov ignore that module.

I added a `.coveragerc` file to the python client directory and explicitly ignored the temp module I create. That's possibly sensitive to changes, but I don't want to ignore things that should not be ignored.

client/python/cryoet_data_portal/Makefile
manasaV3: That sounds good to me! :)

manasaV3 (approved):

PR #1026, opened 2024-08-13
fix: Fix Neuroglancer URL bug

### LINKED ISSUE #1025: Neuroglancer visualization are not getting rendered
**Describe the bug**
The neuroglancer visualization are not getting rendered correctly when the view tomogram button is clicked. 

**To Reproduce**
1. Go to any dataset or runs page
2. Click on view tomogram button

**Screenshots**

Screenshot 2024-08-12 at 8 18 28 PM

**Additional context**

There is an additional `'` in the url [here](https://github.com/chanzuckerberg/cryoet-data-portal/blob/main/frontend/packages/data-portal/app/utils/url.ts#L38). I believe it should be 

```
return `https://neuroglancer-demo.appspot.com/#!${encodeURIComponent(
    config,
  )}`
```

bchu1:
closes #1025

### DIFF
--- a/frontend/packages/data-portal/app/utils/url.ts
+++ b/frontend/packages/data-portal/app/utils/url.ts
@@ -35,7 +35,5 @@ export function createUrl(urlOrPath: string, baseUrl?: string): URL {
 }
 
 export function getNeuroglancerUrl(config: string): string {
-  return `https://neuroglancer-demo.appspot.com/#!'${encodeURIComponent(
-    config,
-  )}`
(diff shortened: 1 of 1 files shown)


ehoops-cz (approved):

manasaV3:
Can we also add a test around how neuroglancer link gets gets generated to future proof it? 😅

bchu1:
> Can we also add a test around how neuroglancer link gets gets generated to future proof it? 😅

Yeah, I'm thinking of adding a test that checks for the JSON parse error on the Neuroglancer page
image

Issue #920, opened 2024-07-24
Integrate all deposition related backend fields


junxini:

manasaV3:
Sample query for getting all datasets that have annotations from a deposition:

```
query MyQuery {  
  datasets(where: {runs: {tomogram_voxel_spacings: {annotations: {deposition_id: {_eq: <deposition_id>}}}}}) {
    id
  }
}
```

PR #1030, opened 2024-08-14
feat: add filter panel to single deposition page

kne42:
#923 

note that applying filters will not filter the results until #920 and will merely update the search params

Screenshot 2024-08-14 at 4 55 35 PM

### DIFF
--- a/frontend/packages/data-portal/app/graphql/getDatasetsFilterData.server.ts
+++ b/frontend/packages/data-portal/app/graphql/getDatasetsFilterData.server.ts
@@ -1,0 +1,57 @@
+import type { ApolloClient, NormalizedCacheObject } from '@apollo/client'
+
+import { gql } from 'app/__generated__'
+import { Datasets_Bool_Exp } from 'app/__generated__/graphql'
+
+const GET_DATASETS_FILTER_DATA_QUERY = gql(`
+  query GetDatasetsFilterData(
+    $filter: datasets_bool_exp,
+  ) {
+    organism_names: datasets(where: $filter, distinct_on: organism_name) {
+      organism_name
+    }
+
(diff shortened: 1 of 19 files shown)


frontend/packages/data-portal/app/graphql/getDatasetsFilterData.server.ts
```diff
@@ -0,0 +1,57 @@
+import type { ApolloClient, NormalizedCacheObject } from '@apollo/client'
+
+import { gql } from 'app/__generated__'
+import { Datasets_Bool_Exp } from 'app/__generated__/graphql'
+
+const GET_DATASETS_FILTER_DATA_QUERY = gql(`
```
bchu1, line 6: Is it possible to just make this a fragment? https://www.apollographql.com/docs/react/data/fragments/ That way any query can add this just by concatenating the string. It should generate the type too, hopefully that would work with `useTypedLoaderData()`?

```
const TEST_FRAGMENT = gql(`
  fragment TestFragment on query_root {
    datasets {
      organism_name
    }
  }  
`)
```

```
useTypedLoaderData<TestFragmentFragment>()
```

kne42: yes i think so. a lot of the codebase could actually benefit from fragments (for example authors)

how important do you think this is? since i have low bandwidth i would prefer to keep this as-is and make the refactor along with the other places we should be using fragments

bchu1: I don't think it's too important! Maybe file a ticket?

bchu1 (approved):

PR #1040, opened 2024-08-19
feat: deposition filter banner

codemonkey800:
#925

- deposition ID query parameter: `deposition-id`
- prop `banner?: ReactNode` to render banner at the top of `<TablePageLayout />`
- implements deposition banner component and renders at top of dataset or run pages

## Demos

https://dev-depo-filter-banner.cryoet.dev.si.czi.technology/

### Dataset Page

image

### Run Page

image

### DIFF
--- a/frontend/packages/data-portal/app/components/DepositionFilterBanner.tsx
+++ b/frontend/packages/data-portal/app/components/DepositionFilterBanner.tsx
@@ -3,7 +3,10 @@ import { Button, Callout } from '@czi-sds/components'
 import { I18n } from 'app/components/I18n'
 import { QueryParams } from 'app/constants/query'
 import { useQueryParam } from 'app/hooks/useQueryParam'
-import { useDepositionHistory } from 'app/state/filterHistory'
+import {
+  useDepositionHistory,
+  useSingleDatasetFilterHistory,
+} from 'app/state/filterHistory'
 import { I18nKeys } from 'app/types/i18n'
 
 interface Deposition {
@@ -18,6 +21,8 @@ export function DepositionFilterBanner({
   deposition: Deposition
   labelI18n: I18nKeys
 }) {
+  const { singleDatasetHistory, setSingleDatasetHistory } =
+    useSingleDatasetFilterHistory()
   const { previousSingleDepositionParams } = useDepositionHistory()
   const [, setDepositionId] = useQueryParam<string>(QueryParams.DepositionId)
 
@@ -40,7 +45,13 @@ export function DepositionFilterBanner({
         </p>
 
         <Button
-          onClick={() => setDepositionId(null)}
+          onClick={() => {
+            setDepositionId(null)
+
+            const nextHistory = new Map(singleDatasetHistory)
+            nextHistory.delete(QueryParams.DepositionId)
+            setSingleDatasetHistory(nextHistory)
+          }}
           sdsStyle="minimal"
           sdsType="secondary"
         >


frontend/packages/data-portal/app/components/Dataset/RunsTable.tsx
```diff
@@ -61,6 +62,24 @@ export function RunsTable() {
     [searchParams, setSingleDatasetHistory],
   )
 
+  const getRunUrl = useCallback(
+    (id: number) => {
+      const url = createUrl(`/runs/${id}`)
+
+      if (deposition) {
+        maybeAddDepositionIdFilter({
```
bchu1, line 70: wonder if we could extract the filter carryover logic separately from the add deposition filter logic? Or maybe just rename this function? Kinda confused me cuz I thought all this function was doing was just tacking on 1 param

frontend/packages/data-portal/app/graphql/getDatasetById.server.ts
```diff
@@ -149,6 +150,11 @@ const GET_DATASET_BY_ID = gql(`
         }
       }
     }
+
+    deposition: datasets(where: { id: { _eq: $deposition_id } }) {
```
bchu1, line 154: is this supposed to be querying `deposition`?

frontend/packages/data-portal/app/components/I18n.tsx
```diff
@@ -26,6 +26,7 @@ export function I18n({ i18nKey, components, linkProps, ...props }: Props) {
         bold: <span className="font-bold" />,
         semibold: <span className="font-semibold" />,
         code: <span className="font-mono" />,
+        dash: <span className="border-b border-sds-gray-600 border-dashed" />,
```
kne42, line 29: do you think we should merge this functionality with `<Link variant="dashed-underlined">`?

frontend/packages/data-portal/app/components/Dataset/RunsTable.tsx
```diff
@@ -61,6 +63,25 @@ export function RunsTable() {
     [searchParams, setSingleDatasetHistory],
   )
 
+  const getRunUrl = useCallback(
+    (id: number) => {
+      const url = createUrl(`/runs/${id}`)
```
bchu1, line 68: nit - since we're only using the path and the query params, do we need `createUrl()`? what if `carryOverFilterParams()` just returned a new `URLSearchParams` and we did ```return `/runs/${id}\` + urlSearchParams.toString()```?

frontend/packages/data-portal/app/routes/datasets.$id.tsx
```diff
@@ -20,6 +21,7 @@ export async function loader({ params, request }: LoaderFunctionArgs) {
 
   const url = new URL(request.url)
   const page = +(url.searchParams.get(QueryParams.Page) ?? '1')
+  const depositionId = +(url.searchParams.get(QueryParams.DepositionId) ?? '-1')
```
bchu1, line 24: the loader function already has a default to `-1`

bchu1 (approved):

bchu1:
Another thought - now that we have the deposition ID in the URL params, should we just consolidate the previous deposition state into the query param?

Janeece (changes requested):
@codemonkey800 I have a few comments:

1. The dashed link in the banner should go to the deposition page
2. "Remove filter" doesnt do anything (should remove filter and banner) -- i assume this is because the deposition ID filter isnt implemented yet, but wanted to call out just  in case
3. Banner is only showing when i set a filter on the dispositions page first. 
    - Banner should appear whenever i go from deposition page > single dataset or deposition page > single dataset > single run
    - banner is tied to the "deposition ID" filter, and no others
    -  lmk if you need a screens recording

Janeece (approved):
@codemonkey800 thank you, it looks great!

PR #1068, opened 2024-08-22
feat: annotation GO ID -> annotation Object ID: add UniProtKB support to annotation Object IDs

daniel-ji:
Renames GO ID field of annotation object (and any relevant references) to Object ID, now supporting both GO IDs and UniProt accessions (IDs). In preparation for new datasets that will be ingested that have UniProt accessions. 

Example Provided value from graphql / api:
`UniProtKB:Q03135`
Which will link as:
`https://www.uniprot.org/uniprotkb/Q03135`

GO ID features are still fully functional.

A question: Are there tests currently that check that the link for an annotation object id works? I can add to any existing ones to also check that UniProt IDs also display & link properly. And do we have any sample data that I can add UniProt IDs to to ensure filtering works as intended? Thanks!


client/python/cryoet_data_portal/src/cryoet_data_portal/data/schema.graphql
```diff
@@ -1194,7 +1194,7 @@ type annotations {
   """
   object_description: String
 
-  """Gene Ontology Cellular Component identifier for the annotation object"""
```
bchu1, line 1197: Are these auto-generated from another source? (Hasura?) May be nice to have a BE person on the PR as well?

bchu1:
Is there ever a situation where an annotation would have both GO and UniProt IDs?

frontend/packages/data-portal/app/components/Run/AnnotationObjectTable/components/InfoLink/InfoLink.tsx
```diff
@@ -0,0 +1,26 @@
+import { Link } from 'app/components/Link'
+import {
+  GO,
+  GO_PATTERN,
+  UNIPROTKB,
+  UNIPROTKB_PATTERN,
+} from 'app/constants/annotationInfoLinks'
+
+export function InfoLink({ id }: { id: string }) {
```
bchu1, line 9: nit - should this be called ObjectIdLink or something, since it's only for displaying object IDs?

frontend/packages/data-portal/app/constants/annotationInfoLinks.ts
```diff
@@ -0,0 +1,6 @@
+export const GO = 'https://amigo.geneontology.org/amigo/term/'
+export const UNIPROTKB = 'https://www.uniprot.org/uniprotkb/'
+
+export const GO_PATTERN = /GO:[0-9]{7}/
```
bchu1, line 4: curious if these regexes need to be so specific? What if we just matched "GO:" and "UniProtKB:"?

client/python/cryoet_data_portal/src/cryoet_data_portal/data/schema.graphql
daniel-ji: @jgadling, Do you know? Thanks!

frontend/packages/data-portal/app/constants/annotationInfoLinks.ts
daniel-ji: These are the same regexes on the backend that ensure the values are correct, so I thought I'd just use the same here. Based on: https://bioregistry.io/registry/go and https://www.uniprot.org/help/accession_numbers

frontend/packages/data-portal/app/components/Run/AnnotationObjectTable/components/InfoLink/InfoLink.tsx
daniel-ji: Changed, thanks!

uermel:
> Is there ever a situation where an annotation would have both GO and UniProt IDs?

This could be the case in a future iteration of the annotation metadata. At this time this is a gapstop fix to allow providing correct metadata for the annotations generated for the ML Challenge.

client/python/cryoet_data_portal/src/cryoet_data_portal/data/schema.graphql
jgadling: Yep, this file is read from the hasura api schema. In order for this change to stick, we'll need a PR for the backend repo along these lines:
https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql

daniel-ji: Sounds good, thanks! I'll go ahead and revert these changes then.

bchu1 (approved):

Issue #747, opened 2024-05-22
Upgrade SDS


codemonkey800:
Upgrade SDS to the latest version and refactor any references based on the [upgrade guide](https://docs.google.com/document/d/1stWFW8lUH43gfP8qVopB-eiV2Vr_4479iqoYR7EmAC4/edit).

PR #1077, opened 2024-08-23
feat: upgrade sds

codemonkey800:
#747

Upgrades SDS to the latest version `20.8.0`. I separated the upgrade into two PRs for better readability: one for the upgrade + refactors, and the 2nd for all of the colors that were grep'd and replaced

## Demo

https://dev-sds-upgrade.cryoet.dev.si.czi.technology

### DIFF
--- a/frontend/packages/data-portal/app/components/Download/ConfigureTomogramDownloadContent.tsx
+++ b/frontend/packages/data-portal/app/components/Download/ConfigureTomogramDownloadContent.tsx
@@ -222,7 +222,7 @@ export function ConfigureTomogramDownloadContent() {
       )}
 
       {multipleTomogramsEnabled && (
-        <Callout intent="warning" className="!w-full !mt-0">
+        <Callout intent="notice" className="!w-full !mt-0">
           {t('annotationsDownloadedFromThePortal')}
         </Callout>
       )}
--- a/frontend/packages/data-portal/app/components/Download/DownloadOptionsContent.tsx
+++ b/frontend/packages/data-portal/app/components/Download/DownloadOptionsContent.tsx
@@ -157,7 +157,7 @@ export function DownloadOptionsContent() {
       <DownloadTabContent />
 
       {multipleTomogramsEnabled && (
-        <Callout intent="warning" className="!w-full">
+        <Callout intent="notice" className="!w-full">
           {t('annotationsDownloadedFromThePortal')}
         </Callout>
       )}


frontend/packages/data-portal/app/components/Filters/DropdownFilterButton.tsx
```diff
@@ -62,8 +62,8 @@ export function DropdownFilterButton({
             'group-hover:!fill-black transition-colors',
             open ? '!fill-black' : '!fill-sds-gray-500',
           )}
-          sdsIcon="chevronDown"
-          sdsSize="s"
```
ehoops-cz, line 66: This size change was intentional?

ehoops-cz (approved):
Looks good to me.  I defer to design on this one.

Thanks for working on this!

frontend/packages/data-portal/app/components/Filters/BooleanFilter.tsx
```diff
@@ -16,7 +16,7 @@ export function BooleanFilter({
   value: boolean
 }) {
   return (
-    <div className={cns('pl-sds-m whitespace-nowrap', styles.boolean)}>
+    <div className={cns('pt-sds-m pl-sds-m whitespace-nowrap', styles.boolean)}>
```
bchu1, line 19: was the switch from left to top intentional?

frontend/packages/data-portal/app/components/Filters/SelectFilter.tsx
```diff
@@ -33,15 +34,19 @@ export function SelectFilter<
 }: {
   className?: string
   details?: ReactNode
-  groupBy?: (option: Value<Option, Multiple>) => string
+  groupBy?: (
+    option: AutocompleteValue<Option, Multiple, DisableClearable, FreeSolo>,
```
bchu1, line 38: is it possible for these to be typed as `SdsOption`?

frontend/packages/data-portal/app/components/Filters/SelectFilter.tsx
```diff
@@ -79,13 +89,11 @@ export function SelectFilter<
...
     if (isArray(value)) {
-      return value.map(convertBaseOptionToSdsOption) as unknown as SdsOption
+      return value.map(convertBaseOptionToSdsOption) as SdsOption
     }
 
     if (value) {
-      return convertBaseOptionToSdsOption(
-        value as Option,
-      ) as unknown as SdsOption
+      return convertBaseOptionToSdsOption(value as Option) as SdsOption
```
bchu1, line 96: is it possible for this function to just return `SdsOption` too?

frontend/packages/data-portal/app/components/Select.tsx
```diff
@@ -91,6 +111,9 @@ export function Select({
       </div>
 
       <InputDropdown
+        // For some reason input dropdown says `className` is not defined even though it is.
```
bchu1, line 114: I did see this https://github.com/chanzuckerberg/sci-components/blob/252c86cbffa7c38b901ac501be5ef72d9998bb45/packages/components/src/core/InputDropdown/style.ts#L49 though if it's just a temporary hack for them that might not be much better

bchu1:
🎉 

Only noticed 2 things from the demo
 - Looks like the font changed?
 - Seems like there's an extra space on the top of the select menus now for some reason

frontend/packages/data-portal/app/components/Filters/SelectFilter.tsx
tihuan: CC: @masoudmanson for help here!

frontend/packages/data-portal/app/components/Download/DownloadModal.tsx
```diff
@@ -106,12 +106,14 @@ function DownloadModalContent() {
         </div>
 
         <div>
-          <ButtonIcon
-            onClick={closeModal}
-            sdsSize="medium"
-            sdsType="tertiary"
-            sdsIcon="xMark"
-          />
+          <button onClick={closeModal} type="button">
```
masoudmanson, line 109: The `ButtonIcon` component has been merged into the `Button` component, you can use it like this:

```js
<Button
  sdsType="primary"
  sdsStyle="icon"
  icon="XMark"
  sdsSize="medium"
  onClick={() => {}}
/>
```

frontend/packages/data-portal/app/components/Filters/SelectFilter.tsx
masoudmanson: @tihuan , we could definitely export some type aliases, but I believe encapsulating those generic parameters with fixed values directly in SDS might not be the best approach.

@bchu1 , does it make sense to define a type alias in your specific context and use it within the component? This way, it remains flexible and context-specific.

```js
// Single Select
type SDSOption = AutocompleteValue<Option, false, false, false>;

// Or maybe a Multi Select
type SDSOptionMultiple = AutocompleteValue<Option, true, false, false>;
```

bchu1: oh yeah, I just meant can we use our existing type alias

frontend/packages/data-portal/app/components/Filters/DropdownFilterButton.tsx
Janeece: @codemonkey800 Use XS -- this is the size used for the minimal dropdown component in [SDS's documentation](https://sds.czi.design/009eaf17b/p/1004b1-dropdown-input/t/page-1004b1-73023306-993ff3-17)

frontend/packages/data-portal/app/components/Filters/BooleanFilter.tsx
Janeece: @codemonkey800 For the type size, this one is a customization from the checkbox component so that it is the same as the filter dropdowns (hierarchy thing). font size should be `fontBodyS`

I believe it should have top padding, but we should also ensure the padding looks even with the other filters in that section (currently too much space between ground truth check box and "available files in the dev link) -- cc @kev-zunshiwang 

Screenshot 2024-08-28 at 2 12 56 PM

frontend/packages/data-portal/app/components/Filters/SelectFilter.tsx
tihuan: Thanks so much for the feedback, @codemonkey800 💡 !

Yeah I don't remember there was a strong reason for SDS to make `name` a required prop in the option and deviate from MUI's default `label` as the required prop. In general it feels like following MUI's default makes sense and when SDS deviates some MUI, it needs to have a strong justification

Actually @sidneymbell ran into the same issue a year ago or so too. So since that's now a recurring pattern, maybe it's worth the breaking change?

@masoudmanson what do you think?

kne42 (commented):
Looks great, thanks for taking on this, it totally would've given me a headache 😅 

Could you please also address the `FIXME`s related to the SDS upgrade? pattern `FIXME:.+sds`

These should be in the `AnnotationTable` and `GroundTruthAnnotationFilter`

It might also be worth doing a generic comment search for SDS-related bugs to see if they're fixed, as well as replacing icons we manually ported over

frontend/packages/data-portal/app/components/Filters/SelectFilter.tsx
masoudmanson: @tihuan Absolutely! I’ll add this to the biweekly agenda so we can discuss the breaking change with the rest of the team and find out if there was a specific reason for making the "name" prop required in the option. 😬🥳

tihuan: Thanks so much, @masoudmanson ! Now thinking about it, I might have been the one who introduced "name" as required years ago, thinking that `label` is not as intuitive as name 🤦 💀 So if no one else objects, we should just go with `label` 😆

masoudmanson: Haha, no worries at all 😯😬

I’ll bring this up in the breaking changes channel to make sure everyone’s aligned on switching from `name` to `label` and `id`. If there are no objections, I’ll get started on a PR to make the update.

tihuan: Thank you so much!! I regret changing it from label to name years ago now 😭 😆

Janeece (changes requested):
@codemonkey800 Thanks for making all of the changes we listed in the design feedback doc -- they look resolved to me (cc @kev-zunshiwang )

**I noticed a couple of items (i'm not sure if they are new or i just missed them before):**

- [ ] On home page - header button has a blue border - please remove for this specific button. 
Screenshot 2024-09-13 at 10 40 31 AM

- [ ] Authors legend tooltip (all locations) -- looks like right size padding is missing or tooltip is too narrow? pls increase tooltip width so it has the right amount of padding
Screenshot 2024-09-13 at 10 53 05 AM

- [ ] Low priority - some links are a darker blue (like the SDS link hover state), while some are brighter (the sds default state color) — they should be the brighter blue as spec'd by SDS. Locations where I see darker color: home page links, ML challenge page links.

Issue #917, opened 2024-07-24
Add deposition related filters on existing pages


junxini:
Add the following filters onto our existing pages to account for the new deposition related work:

Browse all data tab
1. Update existing Author filter to "Dataset Author"
2. Add new "Deposition ID" filter to browse all data tab
3. Add "GO ID" filter (note this might not be shown in the design link)

Single Dataset page
1. Add "Deposition ID" filter, add it under Name/ID filter section
2. Add "GO ID" filter

Single Run page
1. Add "Deposition ID" filter, add it under Name/ID filter section
2. add a new section header "Annotation Metadata" 

Link to designs: https://www.figma.com/design/Rrr54QysCLa14vOnQcmejb/Janeece---Annotation-Depositions?node-id=3381-38903&t=VCWhmasz59xlMrGy-0

PR #1079, opened 2024-08-27
feat: add deposition related filters

kne42:
#917 
- add new filter type: ID Filter which can take in a prefix and validate IDs to use it
- add deposition ID filter to browse all datasets, single dataset, and single run pages
- add gene ontology filter to single dataset page

note: will not actually filter anything since backend is not hooked up yet

Screenshot 2024-08-27 at 3 42 30 PM

live link: https://dev-kira-1.cryoet.dev.si.czi.technology/


frontend/packages/data-portal/app/components/AnnotationFilter/GeneOntologyFilter.tsx
```diff
@@ -1,27 +1,17 @@
...
   return (
-    <MultiInputFilter
-      label={t('goId')}
+    <IDFilter
+      id="go-id-input"
       title={t('filterByGeneOntologyId')}
-      filters={[GO_FILTER]}
+      label={t('goId')}
+      queryParam={QueryParams.GoId}
+      prefix="GO:"
```
bchu1, line 14: could we reference `GO_PREFIX`?

frontend/packages/data-portal/app/components/Filters/IDFilter.tsx
```diff
@@ -0,0 +1,95 @@
...
+  id: string
+  label: string
+  title: string
+  queryParam: QueryParams
+  prefix?: string
+}) {
+  const { t } = useI18n()
+  const [searchParams, setSearchParams] = useSearchParams()
+
+  const paramValue = searchParams.get(queryParam) ?? ''
```
bchu1, line 26: since we're just using `paramValue` in a ternary, do we need the `?? ''`?

frontend/packages/data-portal/app/components/Filters/IDFilter.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [value, setValue] = useState(paramValue)
+
+  const validatingRegex = useMemo(
```
bchu1, line 29: could you add a comment explaining this regex?

frontend/packages/data-portal/e2e/pageObjects/filters/utils.ts
```diff
@@ -71,6 +72,16 @@ export function getExpectedTotalCount({
   )
 }
 
+export function getPrefixedId({
```
bchu1, line 75: nit - better to just have 2 arguments so the types are easier to read

bchu1 (approved):

Janeece (approved):
@kne42 LGTM, thanks!

frontend/packages/data-portal/app/components/Filters/IDFilter.tsx
kne42: this is done so it (and therefore `value`) will be typed as a string always instead of `string | null`

kne42:
@bchu1 could u go over this for re-approval pls since i had to do a pretty heavy refactor to resolve merge conflicts?

frontend/packages/data-portal/app/components/Filters/RegexFilter.tsx
```diff
@@ -0,0 +1,96 @@
+import { useSearchParams } from '@remix-run/react'
+import { useMemo, useState } from 'react'
+
+import { QueryParams } from 'app/constants/query'
+import { useI18n } from 'app/hooks/useI18n'
+
+import { DropdownFilterButton } from './DropdownFilterButton'
+import { InputFilter } from './InputFilter'
+
+export function RegexFilter({
```
bchu1, line 10: Wonder if we should make this component more generic, cuz now we don't have a general component that's just an input field with an "Apply" right? Maybe `regex` should be optional? (could just be a TODO too)

frontend/packages/data-portal/app/components/Filters/IdFilter.tsx
```diff
@@ -0,0 +1,40 @@
+import { useMemo } from 'react'
+
+import { QueryParams } from 'app/constants/query'
+
+import { RegexFilter } from './RegexFilter'
+
+export function IdFilter({
```
bchu1, line 7: since this looks like it's supposed to be filtering CryoET entity prefixes specifically, maybe name it EntityIdFilter or something? 🤷

frontend/packages/data-portal/app/constants/idPrefixes.ts
```diff
@@ -0,0 +1,3 @@
+export enum IdPrefix {
```
bchu1, line 1: Nice

frontend/packages/data-portal/app/components/Filters/IdFilter.tsx
```diff
@@ -0,0 +1,40 @@
...
+  label,
+  title,
+  queryParam,
+  prefix,
+}: {
+  id: string
+  label: string
+  title: string
+  queryParam: QueryParams
+  prefix?: string
```
bchu1, line 18: type this to `IdPrefix`

frontend/packages/data-portal/app/components/Filters/IdFilter.tsx
```diff
@@ -0,0 +1,40 @@
...
+    [prefix],
+  )
+
+  return (
+    <RegexFilter
+      id={id}
+      label={label}
+      title={title}
+      queryParam={queryParam}
+      regex={validatingRegex}
```
bchu1, line 31: should the prop be called `validationRegex` as well?

frontend/packages/data-portal/app/components/Filters/IdFilter.tsx
```diff
@@ -0,0 +1,40 @@
...
+  prefix,
+}: {
+  id: string
+  label: string
+  title: string
+  queryParam: QueryParams
+  prefix?: string
+}) {
+  const validatingRegex = useMemo(
+    () => (prefix ? RegExp(`^(?:${prefix}-)?(\\d+)$`, 'i') : /^\d+$/),
```
bchu1, line 21: since the filter isn't using the capture groups from the regex anymore, is it possible to just remove `?:`?

PR #1085, opened 2024-08-29
feat: Initial support for apiv2

jgadling:
This is a checkpoint: Basic relationships and filters work, but I haven't had a chance to test all of the filter operators and update test infra yet.


andy-sweet (commented):
Generally looks good, and I'm happy that the codegen module didn't need to be changed much.

After setting up an [v2 backend](https://github.com/chanzuckerberg/cryoet-data-portal-backend/tree/main/apiv2), I was able to run some quick manual tests, so agreed this seems to be kinda working. Lots of the list relationships give empty lists, but that could just be due to the local test infra fakes.

With the new version of the backend, can we specify comments/descriptions on the item/list relationships? That was not possible with hasura, so I just had to autogenerate those descriptions with sometimes awkward results (e.g. ` The alignment this per section alignment parameters is a part of`).

client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
```diff
@@ -135,7 +138,7 @@ def get_related_class(self):
         return self.__current_query.get_related_class()
 
     def to_gql(self):
-        return self.__name
+        return strcase.to_lower_camel(self.__name)
```
uermel, line 141: This seems to cause an exception in `strcase`, as `self.__name` is a list, not a string. 

to reproduce: 
```python

client = cdp.Client()
point_annos = cdp.AnnotationFile.find(
    client,
    [
        cdp.AnnotationFile.tomogram_voxel_spacing.run_id == 1000,
        cdp.AnnotationFile.annotation_shape.shape_type._in(["Point", "OrientedPoint"]),  # noqa
    ],
)
```

`TypeError: expected string or bytes-like object`

uermel: Fix seems to be as easy as 

```python
return [strcase.to_lower_camel(n) for n in self.__name]
```

uermel (changes requested):

uermel (approved):

PR #1092, opened 2024-08-31
feat: info panel deposition metadata

codemonkey800:
#918

Implements deposition metadata frontend in the info panel. This only implements the frontend, so it will need to be integrated with the backend as part of #920 

## Demos

image

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/AnnotationOveriewTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/AnnotationOveriewTable.tsx
@@ -135,7 +135,14 @@ export function AnnotationOverviewTable() {
                   <ul>
                     {METHOD_LINKS.map((link) => (
                       <li key={`${link.url}_${link.i18nLabel}_${link.title}`}>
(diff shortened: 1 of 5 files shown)


frontend/packages/data-portal/app/components/Run/AnnotationOveriewTable.tsx
```diff
@@ -37,6 +79,19 @@ export function AnnotationOverviewTable() {
...
+          renderValue: () => (
+            <Link className="text-sds-primary-400" to="/deposition/123">
+              Deposition Name
+            </Link>
+          ),
+        },
+        {
+          label: t('depositionId'),
+          values: ['CZCDP-12345'],
+        },
```
kne42, line 94: can you hide this behind a feature flag?

kne42 (changes requested):
the `MethodLink` component should have two variations as some of its classes differ slightly in `MethodLinksOverview` and `MethodLinksMetadataTable`

kne42 (approved):
looks great! thanks!

PR #1114, opened 2024-09-05
ci: skip e2e if feature disabled

codemonkey800:
Adds a new util function to skip e2e tests if a certain feature is not enabled. This should also fix the prod E2Es that have been failing for a bit


frontend/packages/data-portal/e2e/browseDatasetFilters.test.ts
```diff
@@ -1128,6 +1129,8 @@ test.describe('Browse datasets page filters', () => {
   })
 
   test.describe('Deposition IDs filter group', () => {
+    skipIfFeatureIsDisabled('depositions')
```
bchu1, line 1132: nit - name onlyRunIfEnabled or something like that? The double negatives 😆

bchu1 (approved):

Issue #1005, opened 2024-08-08
Download single tomogram updates + BE integration


junxini:
For single download dialog step 1:
- combine the 2 dropdown options in production into a single dropdown option titled "Select Tomogram"
  - If user selects to download from the table, preselect table tomogram
  - if user selects to download from the header, preselect portal standard
- dropdown option will contain the ID, tomogram sampling, reconstruction method, and post processing information

For single download dialog step 2: 
- add metadata for tomogram id and reconstruction method

Link to design: https://www.figma.com/design/RsjklVi95DFKatU4Uzi2OG/Janeece---Multiple-Tomograms?node-id=582-18772&t=c9A7Y4sN1Zyl5yHr-1

PR #1120, opened 2024-09-07
feat: Implement new single tomogram download flow

bchu1:
#1005 

Does not include changes dependent on `tomogram.is_portal_standard` and `tomogram.is_author_submitted`.

Fixes bug where `<Select>` doesn't close when selecting the already selected option.

### DIFF
--- a/frontend/packages/data-portal/app/hooks/useDownloadModalQueryParamState.ts
+++ b/frontend/packages/data-portal/app/hooks/useDownloadModalQueryParamState.ts
@@ -38,6 +38,10 @@ export function useDownloadModalQueryParamState() {
     QueryParams.AnnotationId,
   )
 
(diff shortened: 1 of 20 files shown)


frontend/packages/data-portal/app/hooks/useDownloadModalQueryParamState.ts
```diff
@@ -191,6 +216,7 @@ export function useDownloadModalQueryParamState() {
     [getPlausiblePayload, plausible, setDownloadParams],
   )
 
+  /** @deprecated */
   const setTomogramConfig = useCallback(
```
ehoops-cz, line 220: nit: setting this one to `setTomogramConfigDeprecated` and the other one to `setTomogramConfig` would save some future work.

ehoops-cz (approved):

Issue #1015, opened 2024-08-09
Update all ID filters to handle both integers and prefix


junxini:
P0: All ID filter fields across the portal to handle both the integer value(e.g. 12345) and Ids with prefix (e.g. DS-12345)

P1: if possible, have the blue tags show the prefix IDs even if users enter the integer value

dependency on #1049

PR #1124, opened 2024-09-10
feat: Id prefixes in filters

ehoops-cz:
Issue #1015 

When filtering by an id, do not include the prefix in the url query params.  Do include the prefix in the display chip for active filters.

Note: Does this need to be behind a feature flag?

id-prefixes

### DIFF
--- a/frontend/packages/data-portal/app/components/Filters/MultiInputFilter.tsx
+++ b/frontend/packages/data-portal/app/components/Filters/MultiInputFilter.tsx
@@ -5,7 +5,7 @@ import { useCallback, useMemo, useState } from 'react'
 import { QueryParams } from 'app/constants/query'
 import { i18n } from 'app/i18n'
 import { cns } from 'app/utils/cns'
-import { extractNumericId } from 'app/utils/idPrefixes'
+import { extractNumericId, isFilterPrefixValid } from 'app/utils/idPrefixes'
 
 import { DropdownFilterButton } from './DropdownFilterButton'
 import { InputFilter } from './InputFilter'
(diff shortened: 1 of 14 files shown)


frontend/packages/data-portal/app/components/Filters/MultiInputFilter.tsx
```diff
@@ -74,7 +76,9 @@ export function MultiInputFilter({
             const value = values[filter.id]
 
             if (value) {
-              prev.set(filter.queryParam, value)
+              // Our filters currently support numeric IDs and use the queryParam as the key
+              // The filter will show the prefix, but we do not need to store it in the query params
+              prev.set(filter.queryParam, extractNumericId(value) ?? '')
```
bchu1, line 81: Does this affect the author filter? Little concerned cuz the component itself sounds more like a generic filter thing but now it seems to do some pretty ID specific stuff. I wonder if we should add a property to `FilterInputData` for a regex or something? similar to what @kne42 did with `<EntityIdFilter>`?

kne42 (approved):

kne42:
oh also, syncd w/ jun and it doesnt need to be behind a feature flag

frontend/packages/data-portal/app/components/Filters/MultiInputFilter.tsx
ehoops-cz: @bchu1 I just changed it to use a `removeIdPrefix` function instead which first checks to see if the queryParam has a prefix and if it doesn't, just return the input.  Are there instances where we'll have prefix-someString instead of prefix-number?  That won't work with this approach.

I could add a prop - it's not my favorite because I find it hard to click through the editor to find where the props get set.  More flexible though.

Thoughts?

bchu1: I'm ok with the regex (maybe add a comment on the top of the component?), just wondering if this breaks the author name filter cuz that's just a string. Or is that a different component?

ehoops-cz: It previously did break the author name, but it no longer breaks it because it skips any fields that don't have prefixes defined.

bchu1 (approved):

PR #1126, opened 2024-09-10
feat: Prep FE codebase for 2 API URLs

bchu1:
https://github.com/chanzuckerberg/cryoet-data-portal/issues/1101

- Added SSM parameters for temporarily storing the V2 endpoint during the migration.
- Separates the V2 graphql-codegen types into a different file.

image
image


frontend/packages/data-portal/app/__generated_v2__/fragment-masking.ts
kne42, line 1: can you add this to the gitignore?

Issue #1118, opened 2024-09-05
Multiple selections in a single filter are not carried over


kandarpksk:
From the all datasets page, select two object names to filter by, say, microtubule and F1-F0 complex. Open any dataset from the results. Note that only microtubule persists as a filter.

kandarpksk:
I still see the bug. Reproduction steps:
1. Go to https://cryoetdataportal.czscience.com/browse-data/datasets?object=F1-F0+complex&object=membrane&page=15
2. Open this result (it's one of the last results): https://cryoetdataportal.czscience.com/datasets/10301?object=F1-F0+complex
3. Notice that only the first filter selection (F1-F0 complex) is carried over (you can see from the link in the previous step itself!). However, membrane should also be a filter selection.

Another case:
1. Go to https://cryoetdataportal.czscience.com/browse-data/datasets?object=endoplasmic+reticulum+membrane&object=F1-F0+complex
3. Open the first result: https://cryoetdataportal.czscience.com/datasets/10301?object=endoplasmic+reticulum+membrane
4. Notice that only the first filter selection (endoplasmic reticulum membrane) is carried over. This is actually wrong, as this dataset does not have that object annotated. Instead, F1-F0 complex should be the only filter selection. cc: @Janeece to comment on this behavior suggestion. Since this is an OR filter, I don't know if we should show the carryover of an object that isn't annotated, or if it should be shown differently, say.

Janeece:
> cc: @Janeece to comment on this behavior suggestion. Since this is an OR filter, I don't know if we should show the carryover of an object that isn't annotated, or if it should be shown differently, say.

@kandarpksk Ahhh this is an interesting case for carryover filters: If a carryover filter tag is not an available option for the child page, does the tag carry over or does it drop off?

I see 3 potential options:
1. It drops off for the child, but if user goes back to parent, it comes back (somewhat aligned to whole filters falling off if they arent valid for the child page)
2. Carry over all filter tags to child, but user can't select it from filter menu on child (so if removed, can't be reapplied from child page)
3. Carry over all filter tags but change styling on in-applicable tag to make it feel inactive (but still removable. (I'm not sure I've seen this pattern)

![IMG_0231 1](https://github.com/user-attachments/assets/d5f91224-6d67-4879-9370-ae6f95e9c3e5)

For now, I lean towards (1) -- maybe (2) -- and waiting to see if there is any confusion/what the confusion is before going through the design exercise needed for or (3). 

[...]

PR #1093, opened 2024-09-02
feat: integrate depositions backend

kne42:
#920 

live link: https://dev-depositions.cryoet.dev.si.czi.technology/browse-data/depositions

Issue #1081, opened 2024-08-28
Clicking on Back/Forward arrows from browser deletes Filters added instead of navigating to corresponding page


lauramarcos:
**Describe the bug**
 if you try adding in filters and clicking back, it removes the filters instead of bringing you to the previous page

**To Reproduce**
1. Go to https://cryoetdataportal.czscience.com/browse-data/datasets
2. Click on any filter from filters panel and add them
3. Click on <Back> on browser

**Screenshots**

https://github.com/user-attachments/assets/4f367837-2ad8-471d-b91b-e4eeaa9fba51

**Desktop (please complete the following information):**
 - OS: Sonoma 14.5
 - Browser:  Chrome

kev-zunshiwang:
Yeah can we change the back button so it doesnt affect the filters? it should go to the previous page

junxini:
sorry your comment got buried in my emails @codemonkey800, and Kevin's response resurfaced it. I agree with Kevin that it would go to the previous page instead of purely removing/adding filters

kev-zunshiwang:
@codemonkey800 - apparently in kandarp's study, when people click into the download dialog and exit, then click the back button, it brings them into the dialog, is this the same issue? Will this PR fix also fix what's happening to the download dialog problem? @kandarpksk @junxini

kev-zunshiwang:
yeah the back and forward button should be only about going to previous pages the user is on, never about going to the previous feature the user was on ---- could we do a bulk fix on where this all show up?

Issue #527, opened 2024-03-01
Filter carryover behavior


junxini:
If filter is applied, carry it over to the next page so that users don't have to reapply.

This applies for the following filters: 
<img width="665" alt="image" src="https://github.com/chanzuckerberg/cryoet-data-portal/assets/98626158/948dd632-111a-4a33-9942-3bcc120bcd44">

Link to design here: https://www.figma.com/file/q6Z394Xy6wUmaXQ9YFjZH7/Kevin---2023-Spillover-Work?type=design&node-id=2411-27140&mode=design&t=DHk4O4QwzFlrJ00A-0

PR #1128, opened 2024-09-11
feat: filter carry over behavior

codemonkey800:
#527 
#1081
#1118

- Refactors the filter history state to query parameter strings 
- Implements remaining carry over behavior for browse back / forward buttons
- Adds E2E tests (some incomplete until #1093 is merged)

## Demo

https://dev-filter-carry.cryoet.dev.si.czi.technology/

### DIFF
--- a/frontend/packages/data-portal/app/state/filterHistory.ts
+++ b/frontend/packages/data-portal/app/state/filterHistory.ts
@@ -47,9 +47,14 @@ export function useDepositionHistory() {
   }
 }
 
+/**
+ * Syncs URL search parameters with global state via `setParams()` function.
+ * This ensures that we can navigate back to the previous page with the same
+ * filters applied.
+ */
 export function useSyncParamsWithState({
   filters,
-  setParams: setHistory,
+  setParams,
 }: {
   filters: readonly QueryParams[]
   setParams(params: string): void
@@ -67,6 +72,6 @@ export function useSyncParamsWithState({
 
     newParams.sort()
 
-    setHistory(newParams.toString())
-  }, [filters, searchParams, setHistory])
+    setParams(newParams.toString())
+  }, [filters, searchParams, setParams])
 }


frontend/packages/data-portal/app/hooks/useFilter.ts
```diff
@@ -125,46 +125,57 @@ export function useFilter() {
...
-        setSearchParams((prev) => {
-          Object.values(QueryParams).forEach((param) => prev.delete(param))
-          prev.delete('page')
-
-          return prev
-        })
+        setSearchParams(
+          (prev) => {
+            Object.values(QueryParams).forEach((param) => prev.delete(param))
+            prev.delete('page')
```
kne42, line 131: wouldn't this already be covered since `QueryParams.Page` would be deleted by the above line?

frontend/packages/data-portal/app/hooks/useFilter.ts
```diff
@@ -125,46 +125,57 @@ export function useFilter() {
...
       updateValue(param: QueryParams, value?: FilterValue) {
         logPlausibleEvent(param, value)
 
-        setSearchParams((prev) => {
-          prev.delete(param)
-          prev.delete('page')
+        setSearchParams(
+          (prev) => {
+            prev.delete(param)
+            prev.delete('page')
```
kne42, line 145:
```suggestion
            prev.delete(QueryParams.Page)
```

frontend/packages/data-portal/app/hooks/useFilter.ts
```diff
@@ -125,46 +125,57 @@ export function useFilter() {
...
 
       updateValues(params: Partial<Record<QueryParams, FilterValue>>) {
         const entries = Object.entries(params) as [QueryParams, FilterValue][]
         entries.forEach(([param, value]) => logPlausibleEvent(param, value))
 
-        setSearchParams((prev) => {
-          prev.delete('page')
+        setSearchParams(
+          (prev) => {
+            prev.delete('page')
```
kne42, line 163:
```suggestion
            prev.delete(QueryParams.Page)
```

kne42 (commented):
if it's an easy fix, would you also be able to address #1118 in this PR?

frontend/packages/data-portal/app/hooks/useFilter.ts
kne42: probably didn't realize it was included in `QueryParams`

kne42 (approved):

frontend/packages/data-portal/app/state/filterHistory.ts
```diff
@@ -59,10 +49,10 @@ export function useDepositionHistory() {
 
 export function useSyncParamsWithState({
   filters,
-  setHistory,
+  setParams: setHistory,
```
bchu1, line 52: missed the rename here

frontend/packages/data-portal/app/state/filterHistory.ts
```diff
@@ -59,10 +49,10 @@ export function useDepositionHistory() {
 
 export function useSyncParamsWithState({
```
bchu1, line 50: could you add some documentation on what this function does? All the props need to be stable right, cuz this is only supposed to run once when the route loads?

bchu1: Also a little confused by the name now 😆 cuz it sounds like it's syncing search params with local state, but the prop `setParams` is actually for local state?

bchu1: Oh nvm about the first comment, now I see it's supposed to keep running when filters change, and I guess `searchParams` changes so it all works 👍

bchu1 (approved):

PR #1164, opened 2024-09-20
docs: Implement SDS colors for dark mode

melissawm:
Implements basic colors to be consistent with SDS.


melissawm:
Sorry, my mistake.

PR #1169, opened 2024-09-20
feat: Migrate and start using tomograms query in UI

bchu1:
#1101 

Also combines the download selector and table queries into 1.

This temporarily enables GQL response caching for the V2 API while there is an unresolved DB connection limit issue in the new API.

Ideally I would migrate first and then connect the UI to the new data later, but desperate times.

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/TomogramTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/TomogramTable.tsx
@@ -14,10 +14,8 @@ import {
   useMetadataDrawer,
 } from 'app/hooks/useMetadataDrawer'
 import { useRunById } from 'app/hooks/useRunById'
-import {
-  metadataDrawerTomogramAtom,
-  Tomogram,
-} from 'app/state/metadataDrawerTomogram'
+import { metadataDrawerTomogramAtom } from 'app/state/metadataDrawerTomogram'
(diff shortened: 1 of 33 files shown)


frontend/packages/data-portal/app/components/Run/TomogramTable.tsx
```diff
@@ -152,29 +153,28 @@ export function TomogramsTable() {
...
+                    className: '!text-sds-body-xxs !h-sds-icon-xl',
+                    startIcon: (
+                      <Icon sdsIcon="Cube" sdsType="button" sdsSize="xs" />
+                    ),
+                  }}
+                  tooltipPlacement="top"
+                  event={{
+                    datasetId: run.dataset.id,
+                    organism: run.dataset.organism_name ?? 'None',
+                    runId: run.id,
```
ehoops-cz, line 172: this one is still snake case?

ehoops-cz (approved):
LGTM, left a question

frontend/packages/data-portal/app/components/Run/TomogramTable.tsx
bchu1: Yeah the `run` query isn't being used yet. The code is in a confusing state rn and the sooner we can finish the migration the better 😭

Issue #1158, opened 2024-09-18
Selector dropdowns have large paddings


bchu1:
![image](https://github.com/user-attachments/assets/acac4cad-5f44-4075-80c3-feeafc0da6e2)
![image](https://github.com/user-attachments/assets/c75431cb-77a3-4e4a-bdd7-a2221a0069f1)

@Janeece said we should remove the extra padding.

FYI @codemonkey800

PR #1181, opened 2024-09-26
feat: Add standard tomogram badge to tomogram selector

bchu1:
#1005

The messed up paddings are tracked in #1158

image

### DIFF
--- a/frontend/packages/data-portal/app/components/TomogramTypeBadge.tsx
+++ b/frontend/packages/data-portal/app/components/TomogramTypeBadge.tsx
@@ -6,11 +6,13 @@ import { Tooltip } from './Tooltip'
 
 export interface TomogramTypeBadgeProps {
   type: 'standard' | 'author'
+  size?: 'default' | 'small'
   showTooltip?: boolean
 }
 
 export function TomogramTypeBadge({
   type,
+  size = 'default',
   showTooltip,
 }: TomogramTypeBadgeProps) {
   const { t } = useI18n()
@@ -18,7 +20,8 @@ export function TomogramTypeBadge({
   const badge = (
     <div
       className={cns(
-        'h-[20px] px-sds-xs py-sds-xxxs !text-sds-body-xxxs text-[#002660] leading-sds-body-xxxs rounded inline-flex bg-[#e9f1ff]',
+        'inline-flex items-center px-sds-xs py-sds-xxxs !text-sds-body-xxxs text-[#002660] leading-sds-body-xxxs rounded bg-[#e9f1ff]',
+        size === 'small' ? 'h-[16px]' : 'h-[20px]',
         showTooltip && 'cursor-pointer',
       )}
     >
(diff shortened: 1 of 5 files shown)


Janeece (changes requested):
Thanks @bchu1 - The tags look good, but the font weights should be regular instead of semibold. This is not a blocker, but a "should have" for consistency. 

- [ ] Should have - input dropdown font weights appear to be all semibold, only the tomogram name (the "label") should be semibold. 
    - Change voxel spacing & tomogram badge to be regular weight so that it will align with the SDS patterns
- [ ] Should have - dropdown menu text font weights appear to be all semibold, only Menu Item name on the selected menu item should be semibold
    - Change all detail text & the tomogram badge to regular weight
    - confirm that for unselected menu items, the tomogram name is also regular weight (in screenshot this case isnt relevant since there is only 1 tomogram)

Screenshot 2024-09-26 at 1 18 55 PM

bchu1:
> Thanks @bchu1 - The tags look good, but the font weights should be regular instead of semibold. This is not a blocker, but a "should have" for consistency.
> 
> * [x]  Should have - input dropdown font weights appear to be all semibold, only the tomogram name (the "label") should be semibold.
>   
>   * Change voxel spacing & tomogram badge to be regular weight so that it will align with the SDS patterns
> * [x]  Should have - dropdown menu text font weights appear to be all semibold, only Menu Item name on the selected menu item should be semibold
>   
>   * Change all detail text & the tomogram badge to regular weight
>   * confirm that for unselected menu items, the tomogram name is also regular weight (in screenshot this case isnt relevant since there is only 1 tomogram)
> 
> Screenshot 2024-09-26 at 1 18 55 PM

Updated the screenshot!

Janeece (approved):
LGTM, thanks @bchu1

frontend/packages/data-portal/app/components/TomogramTypeBadge.tsx
```diff
@@ -6,19 +6,23 @@ import { Tooltip } from './Tooltip'
...
 }: TomogramTypeBadgeProps) {
   const { t } = useI18n()
 
   const badge = (
     <div
       className={cns(
-        'h-[20px] px-sds-xs py-sds-xxxs !text-sds-body-xxxs text-[#002660] leading-sds-body-xxxs rounded inline-flex bg-[#e9f1ff]',
+        'inline-flex items-center px-sds-xs py-sds-xxxs',
+        '!text-sds-body-xxxs text-[#002660] font-normal leading-sds-body-xxxs rounded bg-[#e9f1ff]',
+        size === 'small' ? 'h-[16px]' : 'h-[20px]',
```
ehoops-cz, line 25: 16px is sds-l, there's nothing for 20px though, so not sure if it makes sense to replace one and not the other.

ehoops-cz (approved):

frontend/packages/data-portal/app/components/TomogramTypeBadge.tsx
bchu1: didn't know we could use that with `h-`, yeah if it only works for one I'll just leave it

PR #1190, opened 2024-09-27
feat: Migrate annotations table to V2

bchu1:
#1101


bchu1:
Need `methodLinks` to be populated and a prod API schema sync to continue.

bchu1:
Query is done, filters are done (not tested), UI migration to use new query not done.

Issue #1046, opened 2024-08-20
Download single annotation dialog updates + BE integration


junxini:
https://www.figma.com/design/RsjklVi95DFKatU4Uzi2OG/Janeece---Multiple-Tomograms?node-id=582-18677&t=JVNohTLfE5DZjUsE-0

junxini:
hold off for after design handoff

PR #1191, opened 2024-09-27
feat: Undo changes for transforming annotations and add new callout for mismatched alignments

bchu1:
#1046

Transforming annotations is now out of scope.

Alignment of annotation has not been integrated with the BE yet, since that requires migrating the entire annotations query.

image

image

### DIFF
--- a/frontend/packages/data-portal/app/components/Download/AnnotationAlignmentCallout.tsx
+++ b/frontend/packages/data-portal/app/components/Download/AnnotationAlignmentCallout.tsx
@@ -1,0 +1,60 @@
+import { Callout, CalloutTitle } from '@czi-sds/components'
+
+import { IdPrefix } from 'app/constants/idPrefixes'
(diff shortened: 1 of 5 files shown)


Janeece (commented):
@bchu1 Callout looks good. 

The annotation alignment ID is missing from the top metadata area on both step 1 and step 2 — is this unable to be displayed yet?
Screenshot 2024-09-30 at 4 42 12 PM
Screenshot 2024-09-30 at 4 42 02 PM

bchu1:
> @bchu1 Callout looks good.
> 
> The annotation alignment ID is missing from the top metadata area on both step 1 and step 2 — is this unable to be displayed yet? Screenshot 2024-09-30 at 4 42 12 PM Screenshot 2024-09-30 at 4 42 02 PM

Correct, this requires quite a bit of migration work before we can show that.

Janeece (approved):

PR #1192, opened 2024-09-30
fix: address misc UI/UX bugs

kne42:
see https://docs.google.com/document/d/1jJ56MqvTCBGKbj7XOpdhvGH20MYF0F_ndG2T5jOWQsw/edit?usp=sharing


Janeece (approved):
LGMT, thanks!

frontend/packages/data-portal/app/components/Deposition/MethodLinks/MethodLinksMetadataTable.tsx
```diff
@@ -137,9 +137,13 @@ export function MethodLinksMetadataTable({
         .map((methodData, i) => (
           <div className="flex flex-col gap-sds-xl">
             <MethodSummarySection
-              label={t('methodCount', {
-                value: startCase(converter.toWords(i + 1)),
-              })}
+              label={
+                deposition.annotation_methods.length === 1
```
bchu1, line 141: nit - comment why `1` is the only value that's ignored

frontend/packages/data-portal/app/components/Filters/AnnotationObjectShapeTypeFilter.tsx
```diff
@@ -1,16 +1,25 @@
...
 import { SelectFilter } from './SelectFilter'
 
 function getObjectShapeTypeOptions(
   objectShapeTypes: string[],
+  t: ReturnType<typeof useI18n>['t'],
 ): BaseFilterOption[] {
-  return objectShapeTypes.map((value) => ({ value }))
+  return objectShapeTypes.map((value) => ({
+    label: t(
+      shapeTypeToI18nKey.get(value as ObjectShapeType) ?? ('' as I18nKeys),
```
bchu1, line 19: what does `t('')` actually return? If it errors might be better to put it outside of the `t()`

frontend/packages/data-portal/app/utils/featureFlags.ts
```diff
@@ -7,7 +7,7 @@ export type FeatureFlagEnvironment = typeof process.env.ENV
 export type FeatureFlagKey = 'depositions' | 'multipleTomograms'
 
 export const FEATURE_FLAGS: Record<FeatureFlagKey, FeatureFlagEnvironment[]> = {
-  depositions: ['local', 'dev'],
+  depositions: ['local', 'dev', 'staging', 'prod'],
```
bchu1, line 10: 🎉

bchu1 (approved):

frontend/packages/data-portal/app/components/Filters/AnnotationObjectShapeTypeFilter.tsx
kne42: yeah maybe ill make a util function for it

frontend/packages/data-portal/app/components/Deposition/MethodLinks/MethodLinksMetadataTable.tsx
kne42: design wants if theres only one method for there not to be a "Method One" so we only ignore 1; if there's 0 it wont be executed at all and otherwise all values should display the header

PR #1218, opened 2024-10-04
fix: support changed field name

jgadling:
tomogram.isStandardized -> tomogram.isPortalStandard

NOTE: we should decide how/when we want to deploy this fix.


bchu1 (approved):

bchu1:
I think the actions are failing because it uses the prod schema to compile, so I guess that's expected

PR #1260, opened 2024-10-21
docs: Create simplified data model graphics from script

melissawm:
This is a work in progress as the visual is not very clear right now. I will continue to investigate ways to display the information better.

Preview: TBA


dgmccart:
Yay, glad this work is underway. FYI the new graphic should replace the one here: https://melissawm.github.io/cryoet-data-portal/python-api.html#data-model. The existing graphic on the data organization page will remain the same.

melissawm:
Oh of course, thanks 😅

Issue #624, opened 2024-04-04
Multiple Tomograms Epic


junxini:

bchu1:
BE things needed from the final UX mocks:

1. `run.frames`
2. `run.alignments`
3. `alignment.defocus_value`
4. `alignment.is_canonical`
5. `alignment.alignment_type`
6. `alignment.volume_{x|y|z}_dimension`
7. `alignment.volume_{x|y|z}_offset`
8. `alignment.rotation`
9. `alignment.tilt_offset`
10. `alignment.affine_transformation_matrix`
11. `tomogram.alignment`
12. `tomogram.is_portal_standard`
13. `tomogram.is_author_submitted`
14. `tomogram.related_database_entries`
15. `tomogram.publications`
16. `annotation_file.alignment`
17. Download transformed annotation Portal CLI code snippet
18. Download transformed annotation Python API code snippet
19. New S3 directory structure for downloading all `annotation`s of a `run` (or where the new `s3_prefix` is going to live)
20. New download all annotations Python API code snippet

manasaV3:
@junxini Do we have a decision on the `defocus_value` being a P1 feature for the backend?

manasaV3:
@bchu1  This [document spec](https://docs.google.com/document/d/1InjJKrx78Sv44QguMXR0p1o9Y34oPnMOY9J3aj375ks/edit?usp=sharing) should address most of the fields specified in your comment. I will add the ones it doesn't address below:

> 1. run.frames

This  is currently sourced from `tiltseries.frame_count` this can continue to be sourced from there


The python code snippets will be shared out once that has been finalized.

junxini:
> @junxini Do we have a decision on the `defocus_value` being a P1 feature for the backend?

@uermel could you confirm if this is dependent on frames being imported? If so, frames likely won't get in until early Q4

bchu1:
@codemonkey800

The remaining work is all related to the API migration #1101

- [ ] The run header image and View Tomogram button should point to the `isAuthorSubmitted` tomogram. This is blocked by `isAuthorSubmitted`. 
- [ ] The images and View Tomogram buttons in the tomogram table should point to the `isAuthorSubmitted` tomogram. This is blocked on `isAuthorSubmitted`. https://www.figma.com/design/RsjklVi95DFKatU4Uzi2OG/Janeece---Multiple-Tomograms?node-id=998-14910&node-type=instance&m=dev
[...]

PR #1261, opened 2024-10-21
fix: address multiple tomograms feedback

codemonkey800:
Addresses design feedback from https://github.com/chanzuckerberg/cryoet-data-portal/issues/624#issuecomment-2394596459

## Demo

https://dev-on.cryoet.dev.si.czi.technology/


frontend/packages/data-portal/app/components/Run/TomogramMetadataDrawer.tsx
```diff
@@ -194,7 +196,20 @@ export function TomogramMetadataDrawer() {
           },
           {
             label: t('canonicalStatus'),
-            values: [], // TODO
```
bchu1, line 197: might be good to go through the codebase and search `TODO(bchu)` or something to make sure there isn't something else I missed 😆

bchu1 (approved):

Janeece (approved):
@codemonkey800 - LGTM! A few of notes, but they are either not blockers or I'm not sure if they should be expected as completed:

- P0 - must have - alignment ID field is missing  from Download Single Annotation Step 2 (it was added to step 1)
- P1/P2 - missing Original tomogram field in Annotation info panel
- P2 - nice to have - Tomogram table is still missing max width 
- I assumed that Alignment IDs are just placeholder, so any data value inconsistencies weren't noted

PR #1314, opened 2024-11-06
fix: Handling datetime in client to_dict()

manasaV3:
Calling `to_dict()` method raises serialization errors for objects that have a date time object. Ex: Dataset, Tomogram,.. 

This fix serializes the date time into the iso_format string.

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
@@ -213,19 +213,20 @@ class Model:
             value = getattr(self, k).convert(kwargs.get(strcase.to_lower_camel(k)))
             setattr(self, k, value)
 
-    def _serialize(self, value):
-        if isinstance(value, (datetime, date)):
-            return value.isoformat()
-        return value
-
     def to_dict(self) -> Dict[str, Any]:
         """Return a dictionary representation of this object's attributes"""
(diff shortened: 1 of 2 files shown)


jgadling (changes requested):
This converts datetime objects (which imnsho are correct) to strings. Why do we need to do this?

client/python/cryoet_data_portal/src/cryoet_data_portal/_gql_base.py
```diff
@@ -211,10 +211,19 @@ def __init__(self, client: Client, **kwargs):
...
+            return value.isoformat()
+        return value
+
     def to_dict(self) -> Dict[str, Any]:
         """Return a dictionary representation of this object's attributes"""
         return {k: getattr(self, k) for k in self._get_scalar_fields()}
 
+    def to_json_dict(self) -> Dict[str, Any]:
+        """Return a JSON encoder friendly representation of this object's attributes"""
+        return {key: self._serialize(val) for key, val in self.to_dict().items()}
```
jgadling: Python's `json` module already supplies a way to convert python values to json-compatible values ([docs here](https://docs.python.org/3/library/json.html#json.JSONEncoder.default) and [example implementation here](https://stackoverflow.com/a/35870294) ). Do we need to implement this functionality in our client, or can we leave it to users to work with python's standard library?

jgadling: cc @uermel

PR #1343, opened 2024-11-15
fix: Use correct fieldname in Alignment-TiltSeries relationships.

### LINKED ISSUE #1342: Bug: Alignment-Tiltseries relationship broken in client due to typo.
**Describe the bug**
In the API client models, `Alignment.tiltseries` uses an incorrect field name for `tiltseries_id`.

**To Reproduce**

```python

client = cdp.Client()
tomogram = cdp.Tomogram.get_by_id(client, 771)
print(tomogram.alignment.tiltseries) 
```

`AttributeError: 'Alignment' object has no attribute 'tilt_series_id'`

uermel:
The `Alignment` model was using source field `tilt_series_id` instead of `tiltseries_id`, the `TiltSeries` model was using `tilt_series_id` instead of `tiltseries_id` as a dest field. 

This PR updates both relationships in the python API client. 

closes #1342

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
@@ -70,7 +70,7 @@ class Alignment(Model):
     )
     deposition: Deposition = ItemRelationship("Deposition", "deposition_id", "id")
     deposition_id: int = IntField()
-    tiltseries: TiltSeries = ItemRelationship("TiltSeries", "tilt_series_id", "id")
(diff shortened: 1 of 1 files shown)


jgadling:
This file is codegen'd, so I think we need to figure out where the root cause of this error is coming from 😓

uermel:
> This file is codegen'd, so I think we need to figure out where the root cause of this error is coming from 😓

mmh, that's a shame. this breaks the relationship between tomogram and tiltseries, so it would be great if we could fix it. unfortunately I know too little about where the schema for codegen comes from to investigate this.

PR #1348, opened 2024-11-20
fix: Fixes for relationship field naming.

jgadling:
An alternative fix for #1343 that includes updates to the codegen so we don't lose these changes next time we update the schema.

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
@@ -279,7 +279,7 @@ def _parse_model_field(
     logging.debug("_parse_model_field: %s", field_type)
     model = GQL_TO_MODEL_TYPE.get(field_type.name)
     if model is not None:
-        model_field = to_snake(model)
+        model_field = to_snake(name)
         model_name = _camel_to_space_case(model)
         source_model = GQL_TO_MODEL_TYPE[gql_type.name]
         source_model_name = _camel_to_space_case(source_model)
@@ -290,6 +290,14 @@ def _parse_model_field(
             default_value=f'ItemRelationship("{model}", "{model_field}_id", "id")',
         )
     return None
+
+
+def _get_remote_field_name_by_type(
+    model_cls: GraphQLObjectType, lookup_type: GraphQLObjectType,
+) -> str | None:
+    for name, field in model_cls.fields.items():
+        if field.type == lookup_type:
+            return name
 
 
 def _parse_model_list_field(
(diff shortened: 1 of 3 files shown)


uermel (approved):
Thank you, looks great!

client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
```diff
@@ -292,6 +292,14 @@ def _parse_model_field(
     return None
 
 
+def _get_remote_field_name_by_type(
+    model_cls: GraphQLObjectType, lookup_type: GraphQLObjectType,
+) -> str | None:
```
andy-sweet, line 297:
```suggestion
) -> Optional[str]:
```

Or bump the min version of Python from 3.7 to 3.10, which seems a tad aggressive.

client/python/cryoet_data_portal/src/cryoet_data_portal/_codegen.py
```diff
@@ -306,7 +315,7 @@ def _parse_model_list_field(
     of_model = GQL_TO_MODEL_TYPE.get(of_type.name)
     if of_model is not None:
         source_model = GQL_TO_MODEL_TYPE[gql_type.name]
-        source_field = to_snake(source_model)
+        source_field = to_snake(_get_remote_field_name_by_type(of_type, gql_type))
```
andy-sweet, line 318: Returning `None` from `_get_remote_field_name_by_type` seems unexpected. At a minimum, I'd expect to log an error and return `None`. Alternatively, consider just raising either here or in the function. Continuing does not seem like an option because `ListRelationship(., ., None)` should break at some point right?

andy-sweet: Ignore this, I did not read the following correctly at all!

andy-sweet: I clearly still need to wake up (I was looking at `main` for my second comment), but I think my first comment does stand.

jgadling: No problem - I've added a patch. mind taking a look?

andy-sweet (approved):
Seems good enough to me. The extra logic to extract the field name is a bit unfortunate, but not blocking.

One alternative that came to mind is to define `class Tiltseries` (i.e. to make the model name consistent with the gql name), then define `TiltSeries(Tiltseries)` with a deprecation. But I think I prefer this PR.

I guess that changing the gql name to `tilt_series` is not an option?

jgadling:
We decided to keep the TiltSeries class name to improve compatibility with the previous version of the API client, but if we were to start from scratch, the class would be Tiltseries to reflect the field/type names in the API... so we have a little bit of incompatibility with names now that this PR is trying to resolve. And renaming such important fields in the API at this point isn't worth the hassle imop.

PR #1351, opened 2024-11-22
feat: Support a Client object being shared between threads.

jgadling:
This PR adds the ability for a Client object to be shared between threads. Initially, we created one HTTPTransport object that was used for the lifetime of a Client. These transports are not thread safe and can't handle concurrent access!

This PR creates a single Schema object when a Client is instantiated (this takes ~1/10s so we want to minimize how often we redo this work) and then creates a new Transport for each request, using the Schema object we created earlier. Since each request has its own Transport, it's no longer possible for multiple threads to attempt concurrent access to a Transport.

### DIFF
--- a/client/python/cryoet_data_portal/tests/test_get_client.py
+++ b/client/python/cryoet_data_portal/tests/test_get_client.py
@@ -1,3 +1,5 @@
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
 from cryoet_data_portal import Dataset, Run
 from cryoet_data_portal._client import Client
 from cryoet_data_portal._models import Annotation
@@ -84,3 +86,20 @@ def test_item_relationship_with_missing_id(client: Client):
     assert anno
     assert anno.deposition_id is None
     assert anno.deposition is None
+
+
+# Make sure we can fetch data via a thread pool without errors
+def test_thread_safety(client: Client):
+    def _make_request(client: Client):
+        return Annotation.find(client)
+
+    with ThreadPoolExecutor() as tpe:
+        futures = []
+        threads = 5
+        for _ in range(0, threads):
+            futures.append(tpe.submit(_make_request, client))
+        num_results = 0
+        for item in as_completed(futures):
+            num_results += 1
+            assert len(item.result()) == 6
+        assert num_results == threads
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_client.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_client.py
@@ -27,14 +27,6 @@ class Client:
     """
 
     def __init__(self, url: Optional[str] = None):
-        if not url:
-            url = DEFAULT_URL
-        transport = RequestsHTTPTransport(
-            url=url,
-            retries=3,
-            headers={"User-agent": USER_AGENT},
-        )
-
         with open(
             os.path.join(
                 pathlib.Path(__file__).parent.resolve(),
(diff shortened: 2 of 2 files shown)


uermel (approved):
Woohoo! So much speed!

client/python/cryoet_data_portal/tests/test_get_client.py
```diff
@@ -84,3 +86,20 @@ def test_item_relationship_with_missing_id(client: Client):
...
+        return Annotation.find(client)
+
+    with ThreadPoolExecutor() as tpe:
+        futures = []
+        threads = 5
+        for _ in range(0, threads):
+            futures.append(tpe.submit(_make_request, client))
+        num_results = 0
+        for item in as_completed(futures):
+            num_results += 1
```
andy-sweet, line 103: Not worth updating now after the merge, but I think `ThreadPoolExecutor.map` would be equivalent and much more concise here. I.e. something like

```python
for annos in tpe.map(_make_request, (client,) * threads)
    num_results += 1
    assert len(annos) == 6
```

client/python/cryoet_data_portal/tests/test_get_client.py
```diff
@@ -84,3 +86,20 @@ def test_item_relationship_with_missing_id(client: Client):
     assert anno
     assert anno.deposition_id is None
     assert anno.deposition is None
+
+
+# Make sure we can fetch data via a thread pool without errors
```
andy-sweet, line 91: Just to confirm that the scope of this test is to check that the client can be used from a different thread than it was created in? And not to test concurrent requests on multiple threads, which this test may do, but without guarantees.

If that's the case, then I think you could just do something as simple as

```python
    def _make_request(client: Client):
        return Annotation.find(client)

    with ThreadPoolExecutor(max_workers=1) as tpe:
        future = tpe.submit(_make_request, client)
        assert(len(future.result()) == 6)
```

Or if you also want a pre-condition that the client has alread made one request, then you could do so before submitting one on another thread.

Most of the test feels like its testing `ThreadPoolExecutor` rather the intended behavior.

client/python/cryoet_data_portal/tests/test_get_client.py
```diff
@@ -84,3 +86,20 @@ def test_item_relationship_with_missing_id(client: Client):
...
+
+    with ThreadPoolExecutor() as tpe:
+        futures = []
+        threads = 5
+        for _ in range(0, threads):
+            futures.append(tpe.submit(_make_request, client))
+        num_results = 0
+        for item in as_completed(futures):
+            num_results += 1
+            assert len(item.result()) == 6
```
andy-sweet, line 104: I was so confused about why this was 6 until I remember this is testing against the local infra with fixed data.

client/python/cryoet_data_portal/tests/test_get_client.py
jgadling: The code you posted works fine even with the old client -- The client could always be used from a different thread than it was created in. The problem was concurrent access to the transport, so this test tries to **force** that concurrent access. In practice, it seems to accomplish that since I've never seen this succeed with the old client code. I realize that the test in the PR doesn't *guarantee* that we forced the problem state to happen though -- do you know if there's a way to accomplish that?

andy-sweet: Thanks for explaining.

One way to guarantee it is by implementing our own `Transport` that extends the functionality of `RequestsHTTPTransport` and that also uses concurrency primitives, so that we can control the execution of `connect` and `close` in the test to guarantee concurrent requests. Then providing some way to use that at test time (e.g. by passing through an optional initializer argument), or by monkey-patching it in the test.

If were writing lots of tests that needed these guarantees, it would be worth it. But seems like overkill for this one test, especially when it won't flake as a false positive (only as a false negative with some small probability).

Issue #890, opened 2024-07-17
Allow customized examples for common attributes/methods


kandarpksk:
Documentation for inherited attributes/methods such as [Dataset.get_by_id](https://chanzuckerberg.github.io/cryoet-data-portal/python-api.html#cryoet_data_portal.Dataset.get_by_id) is generic (see linked documentation), with examples for the Run class. This is an issue to cover engineering exploration to enable us to have customized documentation for such attributes and methods.

PR #1369, opened 2024-12-02
fix: Fix generated examples formatting

melissawm:
Before: https://chanzuckerberg.github.io/cryoet-data-portal/api_reference.html#cryoet_data_portal.Annotation.get_by_id

Screenshot 2024-12-02 at 12 48 25 PM

After:

Screenshot 2024-12-02 at 12 49 11 PM

EDIT: Closes https://github.com/chanzuckerberg/cryoet-data-portal/issues/890


melissawm:
Can I merge even with the test failure? Not sure if it's related to this PR

client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
```diff
@@ -47,8 +50,9 @@ class {{ model.name }}(Model):
         """
         Examples:
             Get an {{ model.name }} by ID:
+
             >>> {{ model.model_name_underscores }} = {{ model.name }}.get_by_id(client, 1)
-                print({{ model.model_name_underscores }}.name)
+            >>> print({{ model.model_name_underscores }}.name)
```
manasaV3, line 55: This issue predates the changes in the PR, all models don't have the name field, example: TiltSeries, PerSectionAlignmentParameters, etc. 

A simple workaround for this would be to print id in all cases. Or only print name in the cases where the name is a field of the model.

melissawm: Let me know if this is what you mean!

client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
```diff
@@ -47,11 +50,17 @@ class {{ model.name }}(Model):
         """
         Examples:
             Get an {{ model.name }} by ID:
+
             >>> {{ model.model_name_underscores }} = {{ model.name }}.get_by_id(client, 1)
-                print({{ model.model_name_underscores }}.name)
+            {% if "name" in model.fields %}
```
jgadling, line 55:
```suggestion
            {% if model.fields | selectattr("name", "==", "name") | list | length %}
```

jgadling: This was always evaluating to false. Some models *do* have a name field!

Issue #1379, opened 2024-12-06
Model.get_by_id broken in python API client.


uermel:
**Describe the bug**
Recent updates to the portal python API client's docs have broken the method signatures of the client models, in particular, the `Model.get_by_id` methods do not follow the expected signature. `Run` is used as an example below, but this seems to be true for all models.

Previous and expected signature: 
```python
Run.get_by_id(client: Client, id: int) -> Run
```

New signature: 
```python
Run.get_by_id(**kwargs) -> None
```

This causes breakage with any code that relies on `get_by_id`-methods to accept two positional arguments, as was previously the case. **Even when supplying keyword-arguments, the method does not return anything!**

This is especially insidious because the docs actually show examples using positional arguments, e.g.
```python
    @classmethod
    def get_by_id(cls, **kwargs):
        """
        Examples:
            Get an Run by ID:
            >>> run = Run.get_by_id(client, 1)
                print(run.name)
        """
        super(Run, cls).get_by_id(**kwargs) # <-- Note the missing return statement
```

**To Reproduce**
Fails with `TypeError`:
```python
import cryoet_data_portal as cdp

client = cdp.Client()
run = cdp.Run.get_by_id(client, 6764)
```

Returns `None`:
```python
import cryoet_data_portal as cdp

client = cdp.Client()
run = cdp.Run.get_by_id(client=client, id=6764)
```

**Desktop (please complete the following information):**
 - OS: MacOS 15.1.1
 - Python 3.11.10

PR #1380, opened 2024-12-06
fix: fix argument and return value handling for get_by_id

jgadling:
Fix: #1379 

We broke the interface for `get_by_id` when improving the generated documentation for the client. This fixes the interface and adds a test.

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
@@ -43,14 +43,14 @@ class {{ model.name }}(Model):
     find.__func__.__doc__ = Model.find.__func__.__doc__ + find.__func__.__doc__
 
     @classmethod
-    def get_by_id(cls, **kwargs):
+    def get_by_id(cls, *args, **kwargs):
         """
         Examples:
             Get an {{ model.name }} by ID:
             >>> {{ model.model_name_underscores }} = {{ model.name }}.get_by_id(client, 1)
                 print({{ model.model_name_underscores }}.name)
         """
-        super({{ model.name}}, cls).get_by_id(**kwargs)
+        return super({{ model.name}}, cls).get_by_id(*args, **kwargs)
 
     get_by_id.__func__.__doc__ = (
         Model.get_by_id.__func__.__doc__ + get_by_id.__func__.__doc__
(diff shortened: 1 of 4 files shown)


andy-sweet (approved):
This fixes the issue for me and is probably the smallest possible change to fix, so approved. I would personally prefer that the template uses the exact same signature as the base class (i.e. `def get_by_id(cls, client, id)`), but maybe I'm not following the doc generation code carefully enough.

client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
```diff
@@ -43,14 +43,14 @@ class {{ model.name }}(Model):
     find.__func__.__doc__ = Model.find.__func__.__doc__ + find.__func__.__doc__
 
     @classmethod
-    def get_by_id(cls, **kwargs):
+    def get_by_id(cls, *args, **kwargs):
```
uermel, line 46: Not a blocker, but is there any way we can retain the `client, id` signature here?

It's a bit concerning to me that in the process of improving the web docs we're making IDE-based code inspection worse.

client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
```diff
@@ -43,14 +43,14 @@ class {{ model.name }}(Model):
...
         """
         Examples:
             Get an {{ model.name }} by ID:
             >>> {{ model.model_name_underscores }} = {{ model.name }}.get_by_id(client, 1)
```
uermel, line 50: Also not a blocker, but we should find a way to insert class-specific IDs here. The `id=1` will return `None` for almost every important model in the client, so these are not necessarily intuitive examples. I will create a ticket.

uermel (approved):
Thank you for the quick fix!

client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
uermel: Captured here: https://github.com/chanzuckerberg/cryoet-data-portal/issues/1381

client/python/cryoet_data_portal/src/cryoet_data_portal/templates/Model.jinja2
jgadling: Fair enough - I've updated this PR to make the method signatures more explicit. Mind taking another look?

uermel: Yep, this works for me in PyCharm.

Issue #1374, opened 2024-12-05
Unit tests for simple components


codemonkey800:
Unit tests for simpler components that primarily do one thing and are not built up from other multiple simple components.

- [x] Accordion
- [x] AnnotatedObjectsList
- [x] Breadcrumbs
- [x] CollapsibleList
- [x] CopyBox
- [x] DatabaseEntry
- [x] Drawer
- [x] HeaderKeyPhoto
- [x] InlineMetadata
- [x] KeyPhoto
- [x] MenuDropdown
- [x] NoFilteredResults
- [x] Select

PR #1392, opened 2024-12-12
test: simple component unit tests

codemonkey800:
#1374

- Unit tests for simple components
- Upgrade Remix to fix some type errors with tests
  - I was unable to separate the upgrade into a separate PR due to an error related to the the local eslint plugin package
  - It should be safe to have the upgrade included in this PR as long as the E2Es pass

### DIFF
--- a/frontend/packages/data-portal/e2e/apollo.ts
+++ b/frontend/packages/data-portal/e2e/apollo.ts
@@ -2,7 +2,13 @@
  * Version of `apollo.server.ts` that is safe to use in e2e tests.
  */
 
-import * as apollo from '@apollo/client'
+// For some reason we need to use a `default` export even though the types
+// do not have one. It might be playwright doing something weird with the imports
+// or the apollo types have to be updated to export a default export.
+// TODO remove ts-ignore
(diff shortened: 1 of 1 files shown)


frontend/packages/data-portal/app/types/state.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Type definitions exported from useHydrateAtoms.d.ts
+ */
+
+import { WritableAtom } from 'jotai'
+
+export type AnyWritableAtom = WritableAtom<unknown, any[], any>
```
bchu1, line 7: nit - maybe don't export this since it's only being used here?

frontend/packages/data-portal/app/types/state.ts
```diff
@@ -0,0 +1,8 @@
...
+export type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
```
bchu1, line 8: nit - rename this so that it says that it's the atom with the value or something?

bchu1 (approved):
I feel like a good amount of these can probably be deleted if/when we start using Chromatic

PR #1400, opened 2024-12-13
feat: Create versioned documentation pages

melissawm:
Will require a reorg of the gh-pages branch to work correctly.

This is still in draft as there are design questions around the versioning dropdown. Here's how it looks like now

Screenshot 2024-12-13 at 4 02 24 PM
Screenshot 2024-12-13 at 4 02 28 PM


Janeece:
@melissawm here's my thoughts on what designn might look like:

- reduce the size of the version name in the nav, and make it gray so it doesnt compete with the content
- if you can, have dropdown below the selector (not important to do this though)
- Consider giving versions a comparable name or number and labeling "latest" or "stable" as extra context

Design link: https://www.figma.com/design/3yyR5H3PRJjWVeE4U4rkPr/Doc-Site---Version-Dropdown?node-id=2-4993&t=1RHey8VcOA36OilJ-1

Design suggestions

melissawm:
Here's how that looks:

Screenshot 2025-01-13 at 11 08 50 AM

When we have the version numbers I will definitely add those (this does not depend on design, but on the version name/number selected in the conf.py file.

dgmccart (approved):

melissawm:
Conflicts resolved.

For this to work properly, I would have to go in and reorganize the gh-pages branch. Should this be discussed with eng before doing it? This would take a PR to the gh-pages branch and I would then amend this one PR to include the correct deployment strategy. cc @dgmccart

dgmccart:
My understanding is that the gh-pages branch is only used for the documentation site. If that is the case, I do not have concerns about moving forward without consulting eng. If not, ping in the data portal slack first.

docs/conf.py
```diff
@@ -110,6 +110,19 @@
...
+    "version_dropdown": True,
+    "version_info": [
+        {
+            "version": "dev", # version number or path
+            "title": "Latest",
+            "aliases": [],
+        },
+        {
+            "version": "stable",
+            "title": "Stable",
```
melissawm, line 122:
```suggestion
            "title": "Stable (v3.0)",
```

melissawm:
Merging to test with #1513

PR #1407, opened 2024-12-18
chore: update python matrix

kaloster:
- Updated test matrix - removed Python 3.7 and added 3.11 and 3.12
- Fixed `coverage`

### DIFF
--- a/.github/workflows/client-python.yml
+++ b/.github/workflows/client-python.yml
@@ -48,9 +48,12 @@ jobs:
       - name: Upload Coverage
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/client-python.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10']
```
jgadling, line 25: Can we add 3.11 and 3.12?

jgadling (approved):

PR #1421, opened 2024-12-20
Update release.yml

naihsuanshao:
test if running on different runner would work


naihsuanshao:
Closed as this is for debugging only.

Issue #1375, opened 2024-12-05
Unit tests for for feature components


codemonkey800:
Unit tests for more complex components related to specific features:

- [x] DepositionFilterBanner
- [x] InfoLink
- [x] MLChallengeBanner
- [x] ObjectIdLink
- [x] SurveyBanner
- [x] TablePageLayout

PR #1423, opened 2024-12-21
test: unit tests feature components

codemonkey800:
#1375

Adds unit tests for several feature components. I removed a few items from the original ticket:

1. BrowseDataSearch: would be better as an integration or e2e test, also we're planning on revamping search in the future
2. Table components: would be better as an integration or e2e test

### DIFF
--- a/frontend/packages/data-portal/app/components/I18n.mock.tsx
+++ b/frontend/packages/data-portal/app/components/I18n.mock.tsx
@@ -1,5 +1,10 @@
 import type { I18nProps } from './I18n'
 
+/**
+ * Mock I18n component for rendering span element with i18n key as the content.
+ * Any values are passed in the data-attribute prop in case values need to be
+ * tested.
+ */
 export function MockI18n({ i18nKey, values, linkProps }: I18nProps) {
   return (
     <span
--- a/frontend/packages/data-portal/app/utils/mock.ts
+++ b/frontend/packages/data-portal/app/utils/mock.ts
@@ -7,16 +7,12 @@ export function setMockTime(time: string) {
   jest.useFakeTimers().setSystemTime(new Date(time))
 }
 
-export function getMockUser({
-  hasFakeTimers = false,
-}: {
-  /**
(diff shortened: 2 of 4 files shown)


frontend/packages/data-portal/app/components/I18n.mock.tsx
```diff
@@ -0,0 +1,12 @@
+import type { I18nProps } from './I18n'
+
+export function MockI18n({ i18nKey, values, linkProps }: I18nProps) {
```
bchu1, line 3: why do we need this?

frontend/packages/data-portal/app/utils/mock.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import/no-extraneous-dependencies */
+
+import { jest } from '@jest/globals'
+import userEvent from '@testing-library/user-event'
+
+export function setMockTime(time: string) {
+  jest.useFakeTimers().setSystemTime(new Date(time))
+}
+
+export function getMockUser({
```
bchu1, line 10: comments on when to use this?

bchu1 (approved):

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/components/InfoLink/InfoLink.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { createRemixStub } from '@remix-run/testing'
+import { render, screen } from '@testing-library/react'
+
+import { NCBI, OBO, WORMBASE } from 'app/constants/datasetInfoLinks'
+
+import { InfoLinkProps } from './InfoLink'
+
+async function renderInfoLink({ id, value }: InfoLinkProps) {
+  const { InfoLink } = await import('./InfoLink')
```
kne42, line 9: why are we doing a non-global import?

PR #1424, opened 2024-12-23
feat: Add deposition to V2 run query

bchu1:


frontend/packages/data-portal/app/routes/runs.$id.tsx
```diff
@@ -47,17 +47,23 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
     getRunById({
       id,
       annotationsPage,
-      depositionId,
+      depositionId: Number.isNaN(depositionId) ? undefined : depositionId,
       client: apolloClient,
       params: url.searchParams,
     }),
-    getRunByIdV2(apolloClientV2, id, annotationsPage, url.searchParams),
+    getRunByIdV2(
```
kne42, line 60: unrelated nit but would be nice if we used keyword params here so it's more clear what things are for

kne42 (approved):
LGTM!

Issue #1232, opened 2024-10-10
Migrate API for Single Run Page


kne42:
- [x] Query conversion
- [x] Diff logging
- [ ] Refactoring

PR #1433, opened 2024-12-24
feat: Migrate annotations table to V2 data

### LINKED ISSUE #581: Paginate the annotation table on annotation shape type
If the files within an annotation have different shape types, they should be rendered on separate rows.

We'll need to handle pagination a bit carefully. For any given page, the table should have at most 20 annotations on the table. Assume we query the backend and get 20 annotations. If all 20 have files with the same shape type, then the table will render correctly

However, if the first 7 annotations has all 3 files each with different shape types, then we'll only be able to render 20 out of the 21 files using the existing pagination logic. Additionally, the remaining the 13 annotations may never get rendered because when the user clicks on the next page, it paginates only on the `annotations` query and not the `files` query

One approach is to query all of the annotation and file data and handle pagination on the frontend server, but does not feel very ideal because we need to query every single annotation and file for a run which may never get visited if the user doesn't go through each page

### LINKED ISSUE #1020: Migrate paginated query for annotations table to new API
Currently paginating by filtering out .zarr files, paginate on `annotation_shape_types` when available.

resolving temporary solution implemented for #581 

dependency on #936

### LINKED ISSUE #1373: Alignment Id specified incorrectly in the annotation download modal
## Description

As seen in the image below, there are two different alignment ids specified in the download modal. 

The one that has been underlined, has the same value as the annotation id. However the alignment id in copyable text holder seems to have the correct value. 

Image

https://cryoetdataportal.czscience.com/runs/249?download-step=configure&annotation_id=15260&ref-tomogram=undefined&object_shape=Point&file-format=ndjson&annotation_name=41349+fatty+acid+synthase+complex

bchu1:
closes #1373
closes #1020
closes #581
#1232

 - Fixes bug where wrong alignment IDs were being displayed in download modals.
 - Fixes bug where annotation name in sidebar didn't match name in table.
 - Updates most places to use `Annotation_File_Shape_Type_Enum` values from V2 directly.
 - We should migrate the query params for annotation shape selection to just be the annotation shape ID (instead of annotation ID plus shape type) but this PR doesn't do that.


frontend/packages/data-portal/app/components/Download/DownloadOptionsContent.tsx
```diff
@@ -115,16 +119,10 @@ export function DownloadOptionsContent() {
           value={getTomogramName(referenceTomogram)}
         />
       )}
-      {annotationToDownload && tomogramToDownload?.alignment && (
```
rainandbare, line 118: We don't need to check the tomogramToDownload?.alignment anymore?

frontend/packages/data-portal/app/components/Download/DownloadOptionsContent.tsx
```diff
@@ -163,11 +161,16 @@ export function DownloadOptionsContent() {
 
       <DownloadTabContent />
 
-      {annotationToDownload !== undefined && tomogramToDownload?.alignment ? (
+      {annotationFileAlignmentId !== undefined ? (
         <AnnotationAlignmentCallout
-          alignmentId={tomogramToDownload.alignment.id}
```
rainandbare, line 168: Oh I see! re: `tomogramToDownload.alignment.id`

frontend/packages/data-portal/app/components/Download/ConfigureDownloadContent.tsx
```diff
@@ -9,13 +9,16 @@ import { ConfigureTomogramDownloadContent } from './ConfigureTomogramDownloadCon
...
-  const { annotationName, annotationId, objectShapeType } =
+  const { annotationName, annotationId, fileFormat, objectShapeType } =
     useDownloadModalQueryParamState()
-
-  const { annotationToDownload, datasetTitle, runName, objectName } =
+  const { annotationShapeToDownload, datasetTitle, runName, objectName } =
     useDownloadModalContext()
 
+  const annotationFileAlignmentId =
+    annotationShapeToDownload?.annotationFiles.edges.find(
```
rainandbare, line 18: is this something we want to deduplicate?

frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
```diff
@@ -35,32 +38,39 @@ import {
...
+      methodLinks: {
+        edges: [],
+      },
+      methodType: Annotation_Method_Type_Enum.Manual,
+      objectId: '',
+      objectName: '',
+      releaseDate: '',
+    },
+    shapeType: Annotation_File_Shape_Type_Enum.Point,
+  }),
```
rainandbare, line 72: Such a different shape!

frontend/packages/data-portal/app/graphql/common.ts
```diff
@@ -209,7 +208,7 @@ export function getDatasetsFilter({
     where.runs.annotations ??= {}
     where.runs.annotations.annotationShapes = {
       shapeType: {
-        _in: objectShapeTypes as Annotation_File_Shape_Type_Enum[], // TODO(bchu): Remove typecast.
+        _in: objectShapeTypes,
```
rainandbare, line 211: nice

rainandbare (approved):

frontend/packages/data-portal/app/components/Download/DownloadOptionsContent.tsx
bchu1: Yeah this code was super messed up. It was never supposed to display the `alignmentId` of the tomogram (also this condition never evaluated to `true` because it's impossible to download a single annotation and single tomogram at the same time 😭).

frontend/packages/data-portal/app/components/Download/DownloadOptionsContent.tsx
bchu1: Yeah this condition was also just never evaluating to true (effect was that on this page you'd always see the truncated version of the callout)...

frontend/packages/data-portal/app/components/Download/ConfigureDownloadContent.tsx
bchu1: this should just produce on `alignmentId` at the end (technically all the files rn should have the same `alignmentId` but in a future world maybe it won't be)

bchu1:
@rainandbare Thanks for looking through!

PR #1451, opened 2025-01-09
chore: Add test that filters on deposition page at least render

bchu1:


frontend/packages/data-portal/e2e/depositionFilters.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    filtersPage = new FiltersPage(page)
+    filtersActor = new FiltersActor(filtersPage)
+
+    await goTo(page, SINGLE_DEPOSITION_URL)
+  })
+
+  test('Organism name filter renders', async () => {
+    await filtersActor.addSingleSelectFilter({
+      label: 'Organism Name',
+      value: 'Homo sapiens',
```
ehoops-cz, line 22: These should have some sort of "expect" right?

bchu1: I'm just testing that the user can click on them. Didn't want to make any assertions on the resulting data just yet.

Issue #168, opened 2023-11-22
[staging] Tooltips overlap when open one after another


lauramarcos:
**Steps to Reproduce**

1. Hover on a tooltip from Annotated Objects
2. Hover on another tooltip while pop up is still opened

**Actual Results**
pop ups overlap eachother

**Expected Results**
poop ups shouldnt overlap, they should close when a next one is opened

https://github.com/chanzuckerberg/cryoet-data-portal/assets/63799148/dc409003-6fca-437b-8681-576836c49a55

melissawm:
Right now, it seems like the tooltips don't overlap anymore, but there are two pop-ups appearing at each tooltip:

For example, at https://cryoetdataportal.czscience.com/datasets/10440:
![Screenshot 2025-01-13 at 2 32 17 PM](https://github.com/user-attachments/assets/3addbfb4-852f-4caa-9ce0-0d299d44cbbe)

melissawm:
Looks like this is fixed, please reopen if that's not the case.

PR #1460, opened 2025-01-13
fix: Remove duplicated tooltip from Annotated Objects list

melissawm:
Closes #https://github.com/chanzuckerberg/cryoet-data-portal/issues/168

Before:
Screenshot 2025-01-13 at 4 59 17 PM

After:
Screenshot 2025-01-13 at 5 00 26 PM


rainandbare (approved):
🥳 WAHOO! Great fix.

 I ran `pnpm lint:prettier:fix` to get the robots to fix the problems on their own. The front end tests will still likely fail but Bryan hopefully has a PR that will fix that coming in this morning.

PR #1466, opened 2025-01-14
fix: Fix deposition filter on run page

bchu1:
I switched it to using dataset deposition when it should've been annotation deposition.


frontend/packages/data-portal/app/graphql/getDatasetByIdV2.server.ts
```diff
@@ -223,9 +223,8 @@ function getRunFilter(
   // Deposition filter:
   const depositionId = +(filterState.ids.deposition ?? Number.NaN)
   if (!Number.isNaN(depositionId) && depositionId > 0) {
-    where.dataset = {
-      depositionId: { _eq: depositionId },
-    }
+    where.annotations ??= {}
```
kne42, line 226: woah, did not know this was a possible syntax, that's cool

why is it needed instead of a regular equals since it doesn't look like `where.annotations` is set elsewhere? (I'm assuming it's equivalent to `where.annotations = where.annotations ?? {}`)

kne42 (approved):

PR #1467, opened 2025-01-15
feat: Diff logging for deposition page

bchu1:
Lot of it is very similar to the diff logger for datasets page.

Removes hack from datasets diff logger (bug now fixed).

Removes `annotationSoftware` and `datasetPublications` from the query (not used anywhere).

Split the annotation methods fields into 2 (1 for the aggregates per `annotation_method`, 1 for the combination of all `annotation_method`s and `method_links`).

### DIFF
--- a/frontend/packages/data-portal/app/graphql/getDatasetsDiffer.ts
+++ b/frontend/packages/data-portal/app/graphql/getDatasetsDiffer.ts
@@ -63,16 +63,11 @@ export function logIfHasDiff(
     delete group.count
   }
   v2.distinctReconstructionSoftwares.aggregate =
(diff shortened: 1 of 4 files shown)


kne42:
What's the reasoning behind doing the logging this way instead of the pattern followed in the API migration regression tests?

bchu1:
> What's the reasoning behind doing the logging this way instead of the pattern followed in the API migration regression tests?

Mostly due to the number of permutations that these queries can have with the filters and all (though I did realize we don't actually get that much prod traffic 😆 so maybe the data I'm getting is not as great as I thought).

ehoops-cz (approved):

PR #1475, opened 2025-01-16
chore: Remove portal FAQ, point to docs site

### LINKED ISSUE #1462: docs: Remove Portal site faq.mdx
In lieu of a page on the Portal site, we opted to migrate the FAQ to the docsite last year. We have only been updating the docsite FAQ file, so the [Portal site FAQ file](https://github.com/chanzuckerberg/cryoet-data-portal/blob/main/website-docs/faq.mdx) is out of date. We'd like to remove the Portal site FAQ file from the repo to keep our repo clean, but we need to check that all links have been updated to point to the docsite FAQ and implement a redirect from the old URL: https://cryoetdataportal.czscience.com/faq to the docsite FAQ: https://chanzuckerberg.github.io/cryoet-data-portal/cryoet_data_portal_docsite_faq.html

melissawm:
I believe this is all we need, but will check the tests here on the PR and adjust as needed.

Closes #1462 

cc @dgmccart


melissawm:
I suspect the linting failure is due to the line being too long but not sure how to break it up. Help is appreciated!

frontend/packages/data-portal/app/utils/url.test.ts
```diff
@@ -8,8 +8,8 @@ describe('utils/url', () => {
       expect(isExternalUrl('https://www.google.com')).toBe(true)
     })
 
-    it('should return true for non-external URLs', () => {
-      expect(isExternalUrl('/faq')).toBe(false)
+    it('should return true for external URLs', () => {
+      expect(isExternalUrl('https://chanzuckerberg.github.io/cryoet-data-portal/cryoet_data_portal_docsite_faq.html')).toBe(true)
```
rainandbare, line 12: You were right that the linter did not like this long line! 

However I think this test is meant to check that the isExternalUrl function is working so let's change it to this:

```suggestion
   it('should return true for non-external URLs', () => {
      expect(isExternalUrl('/competition')).toBe(false)
    })
```

rainandbare (approved):
I think the fix to the test should do the trick!

rainandbare:
Still angry at us! 👀

PR #1477, opened 2025-01-17
fix: correct alignment id in download modal

### LINKED ISSUE #1373: Alignment Id specified incorrectly in the annotation download modal

kne42:
closes #1373 

before:
Screenshot 2025-01-16 at 5 13 04 PM

after:
Screenshot 2025-01-16 at 5 13 30 PM

### DIFF
--- a/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
+++ b/frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
@@ -345,8 +345,9 @@ export function AnnotationTable() {
                     datasetId: run.dataset.id,
                     runId: run.id,
                     annotationId: annotation.id,
-                    referenceTomogramId: tomograms.find(
-                      (tomogram) => tomogram.isPortalStandard,
+                    referenceTomogramId: (
+                      tomograms.find((tomogram) => tomogram.isPortalStandard) ??
+                      tomograms.at(0)
                     )?.id,
                     objectShapeType: annotation.shape_type,
                     fileFormat: annotation.format,
--- a/frontend/packages/data-portal/app/components/Download/ConfigureDownloadContent.tsx
+++ b/frontend/packages/data-portal/app/components/Download/ConfigureDownloadContent.tsx
@@ -10,11 +10,20 @@ import { ConfigureTomogramDownloadContent } from './ConfigureTomogramDownloadCon
 export function ConfigureDownloadContent() {
   const { t } = useI18n()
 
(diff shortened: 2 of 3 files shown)


frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
```diff
@@ -345,8 +345,9 @@ export function AnnotationTable() {
                     datasetId: run.dataset.id,
                     runId: run.id,
                     annotationId: annotation.id,
-                    referenceTomogramId: tomograms.find(
-                      (tomogram) => tomogram.isPortalStandard,
+                    referenceTomogramId: (
+                      tomograms.find((tomogram) => tomogram.isPortalStandard) ??
+                      tomograms.at(0)
```
bchu1, line 350: I forget, is every run guaranteed to have a portal standard tomogram (if it has any at all)?

bchu1 (approved):

frontend/packages/data-portal/app/components/Run/AnnotationTable.tsx
kne42: no idea, @uermel do you know?

uermel: No, not all of them.

bchu1: So it's possible for a run to have non-zero number of tomograms, and none of them are portal standard?

bchu1: Thanks! So if a user is downloading an annotation from a run and there is no portal standard tomogram in that run, should we not show "Alignment ID: AL-XXXXX" in the download modal? (actually, if there's no portal standard tomogram does that also mean there's no annotations? 🤔)

bchu1: I just realized, this is the modal for downloading an annotation. It should be showing the alignment ID of the `annotationFile` that's being selected by the file type right?

PR #1499, opened 2025-01-21
feat: Migrate rest of dataset page to use V2 data

### LINKED ISSUE #1231: Migrate API for Single Dataset Page
- [x] Query conversion
- [x] Diff logging
- [ ] Refactoring

bchu1:
closes #1231

 - Now that `__typename` has been fixed by the BE, start querying for it so we have an easy way to distinguish between Hasura and Platformics responses (Platformics `__typename`s are capitalized).
 - Platformics dates add a timezone suffix, so remove those from all date fields.
 - Also fix bugs where table values were being incorrectly split into new lines by commas.


frontend/packages/data-portal/app/components/Dataset/DatasetHeader.tsx
```diff
@@ -35,21 +35,21 @@ export function DatasetHeader() {
         </div>
       }
       breadcrumbs={<Breadcrumbs variant="dataset" data={dataset} />}
-      lastModifiedDate={dataset.last_modified_date ?? dataset.deposition_date}
```
rainandbare, line 38: Don't need the deposition_date back up anymore?

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/SampleAndExperimentConditionsTable.tsx
```diff
@@ -89,3 +110,7 @@ export function SampleAndExperimentConditionsTable({
     />
   )
 }
+
+function isV2Dataset(dataset: DatasetType | Dataset): dataset is Dataset {
+  return dataset.__typename === 'Dataset'
+}
```
rainandbare, line 116: Is this just temporary? I am wondering why you have it defined in two places (also in DatasetMetadataTable)

rainandbare (approved):
LGTM! I was able to run it locally!

frontend/packages/data-portal/app/components/Dataset/DatasetHeader.tsx
bchu1: `lastModifiedDate` is required in V2! https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/apiv2/graphql_api/schema.graphql#L1848

frontend/packages/data-portal/app/components/Dataset/SampleAndExperimentConditionsTable/SampleAndExperimentConditionsTable.tsx
bchu1: Yeah, hopefully I can delete this before I go on leave 😬 Figured it would be easier to just have it twice instead of creating more files that I'm just going to delete

PR #1501, opened 2025-01-22
fix: More adjustments to run query differ

bchu1:


frontend/packages/data-portal/app/graphql/getRunByIdDiffer.ts
```diff
@@ -348,15 +352,19 @@ export function logIfHasDiff(
...
+                primaryAuthorStatus: author.primary_author_status,
+                correspondingAuthorStatus: author.corresponding_author_status,
+                name: author.name,
+                email: author.email,
+                orcid: author.orcid,
+              },
+            }))
+            .sort((authorA, authorB) =>
+              authorA.node.name.localeCompare(authorB.node.name),
+            ),
```
rainandbare, line 367: We are sorting it in two places to compare v1 to v2?

rainandbare (approved):

frontend/packages/data-portal/app/graphql/getRunByIdDiffer.ts
bchu1: Exactly, so for fields where both V1 and V2 have non-deterministic ordering, we have to sort both of them (sometimes when it makes sense from the user's perspective I'll just add sorting to the GQL query, but for this one it'll prolly just make things more confusing if I do that bc we're sorting it again in the `<AuthorList>` component)

PR #1507, opened 2025-01-23
fix: Fix intermittent issue where Remix does not replace initial DOM with hydrated version

bchu1:
https://czi-sci.slack.com/archives/C05S98P75NW/p1737656647158829

Not sure why the actual UI bug occurs (sounds like a race condition in Remix?) but fixing the hydration errors (lot of something cannot be descendent of `<p>`) should prevent the wrong SSR HTML from being generated in the first place.


rainandbare (approved):
Thank you for this quick fix Bryan!

PR #1513, opened 2025-01-24
[LARGE DIFF] Set up versioned pages

melissawm:
This PR reorganizes the `gh-pages` branch of the repo to support multiple versions of the documentation site.

It moves all html pages to be in folders corresponding to specific versions (I used v3.0 to match the data model but happy to rename to the correct version number in case I am mistaken). It also includes redirects to stable for a few select pages of the original site that are linked from both the main cryoET data portal and maybe from external pages, namely the index page, the FAQ and the Python API documentation page.

This PR does not affect the client deployment, and should be merged before #1400 .

cc @dgmccart


rainandbare (approved):
Looks good to me! No content changes and correct loading will be checked in #1400. Thanks Melissa!

PR #1518, opened 2025-01-24
fix: Ignore __typename in diff logs

bchu1:
Due to reenabling `__typename` in a previous PR.


rainandbare (approved):
Are we hoping that this will fix the e2e test?

bchu1:
> Are we hoping that this will fix the e2e test?

no, there's nothing we can do on our side to fix the e2e test, this is just so that the diffs aren't going crazy thinking every response has a diff due to the `__typename` fields

PR #1532, opened 2025-01-25
feat: Disable API migration test

bchu1:
Temporarily disable this so that we can test running the prod deploy directly from the main branch.

This is an attempt to fix the issue of prod pods not properly deploying.


manasaV3 (approved):
yolo!!

PR #1553, opened 2025-01-28
chore: Skip failing test

bchu1:
Some tests are going to fail due to the migration because V1 data (and sometimes V1 queries) don't match with V2 all the time. It's going to take considerable effort to migrate the tests over. Also doing prod deploys off the branch was causing issues so we need to get main working again for prod deploys to resume.

### DIFF
--- a/.github/workflows/prod-deploy.yml
+++ b/.github/workflows/prod-deploy.yml
@@ -23,7 +23,7 @@ jobs:
     name: find release PR
     runs-on: ubuntu-latest
     if: github.repository == 'chanzuckerberg/cryoet-data-portal'
-    # needs: run-staging-e2e-test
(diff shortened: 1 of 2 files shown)


rainandbare (approved):
On local I see one more test failing at 
`[webkit] › downloadDialog.test.ts:593:7 › downloadDialog › Single Run › Download Tomogram › should change file type`

bchu1:
> On local I see one more test failing at `[webkit] › downloadDialog.test.ts:593:7 › downloadDialog › Single Run › Download Tomogram › should change file type` but GHA seems happy!

Bleh, maybe I'll have to disable that one too. I wouldn't be doing this if my wife weren't about to give birth at any moment.. 😬

PR #1559, opened 2025-01-29
fix: Stable version redirect

melissawm:
I didn't realize the version number and aliases should be switched in conf.py. This is not urgent but should fix future builds.

Follow-up to #1558


rainandbare (approved):

melissawm:
Confirmed that the dropdown is now working so this is the correct fix, thank you @rainandbare !

PR #1566, opened 2025-01-30
feat: Complete deposition page migration to V2

### LINKED ISSUE #1230: Migrate API for Single Deposition Page
- [x] Query conversion
- [x] Diff logging
- [ ] Refactoring

bchu1:
closes #1230

 - Migrates more hardcoded types to V2 enums.
 - The V2 query for `methodLinks` is much more complicated than in V1, which used `distinct` to query `annotations` and then assumed that all `methodLinks` were the same for the same `annotationMethod` (turned out to not actually be the case, but whatever). V2 doesn't have `distinct`, only `groupBy`, which will return a separate object for each array element of `methodLinks`, so we have to compress them in the FE.


frontend/packages/data-portal/app/constants/methodTypes.ts
```diff
@@ -2,20 +2,13 @@ import { Annotation_Method_Type_Enum } from 'app/__generated_v2__/graphql'
...
-/** @deprecated */
-export const methodLabels: MethodTypeLabels = {
-  automated: 'automated',
-  hybrid: 'hybrid',
-  manual: 'manual',
-}
+export const METHOD_TYPE_ORDER = [
+  Annotation_Method_Type_Enum.Hybrid,
+  Annotation_Method_Type_Enum.Automated,
+  // TODO: Confirm with PM where this should be when BE actually starts returning Simulated.
```
rainandbare, line 8: When will it start?

rainandbare (approved):

PR #1576, opened 2025-02-03
feat: Migrate missing field from depositions query

bchu1:
#1229


rainandbare (approved):
Seems okay to me - re ran the dev build. 

EDIT: did not work - this is the message 
`2.127  ERR_PNPM_LOCKFILE_BREAKING_CHANGE  Lockfile /app/pnpm-lock.yaml not compatible with current pnpm`

hmm maybe that lock file did update for a reason. Try a `pnpm install` ?

PR #1579, opened 2025-02-04
fix: Hold pnpm at 8.10.5 to fix deploys

bchu1:
Grabbing the latest pnpm version no longer supports our `LockfileVersion`. Upgrading our lockfile to latest is causing the build to freeze at `prune --prod`.

Also pork barrel some optional chaining to more safely access Platformics's `aggregate` arrays.


kaloster (approved):
🙌

PR #1581, opened 2025-02-04
fix: Update font to intended Inter font

rainandbare:
I noticed that though the CSS said that Inter was the intended font it was not loaded. It does change the look and feel of the site a bit so I wanted give folks a heads up. 

In this PR:
Add Inter font in header


bchu1:
🙌

bchu1 (approved):

PR #1582, opened 2025-02-04
feat: Stop querying V1

### LINKED ISSUE #1547: Delete V1 code (to enable removing BE removal of V1)

bchu1:
closes #1547

Not trying to change the shape of the object returning by the loader too much because there's a bug where if the loader return value changes in a non-backwards-compatible way, the app will be broken for a short period of time (?). Hopefully someone can figure that out and then update the loaders to remove the `v2` properties.

Not deleting the differs so we can still reference them while migrating the tests.


bchu1:
> Do you need any help with the backwards compatible loaders situation?

Not urgent (and actually doesn't come up very often, esp now that we should just be making 1 query per page), but would be nice for someone to figure out eventually 😆

PR #1591, opened 2025-02-06
fix: Runs table View Tomogram button styling

bchu1:
New (old) font is slightly bigger, needs more space.


rainandbare (approved):
Thank you!

PR #1601, opened 2025-02-10
chore: Enable Chromatic

bchu1:
#1608 #1609

 - Tests that are only run during the prod deploy have been deleted. These tests are now guaranteed to fail whenever a diff is made between deploys because their diffs won't be updated by PRs. This includes:
   - Firefox and Webkit browsers.
   - Prod environment tests.
 - Migrates our smallest test (carryOver.test.ts) to Chromatic.
 - Chromatic automatically takes a screenshot at the end of every test, so the final assertion of each test should mostly be deletable. In the future we should disable the automatic screenshotting so that we can take targeted screenshots of specific DOM elements, which will require adding a manual call to `takeSnapshot()` back to the end of every test.
 - Fixes some unit tests (something updated?).

To view diffs on PRs, click the "UI Tests" job.


frontend/packages/data-portal/package.json
```diff
@@ -9,6 +9,7 @@
     "build:remix": "remix build",
     "clean": "rm -rf build public/build .cache/ node_modules/.cache/ app/__generated__/",
     "clean:modules": "rm -rf node_modules",
+    "chromatic": "chromatic --playwright --project-token=chpt_fda0cc83112e755",
```
tihuan, line 12: since chromatic project token is passed in its GHA directly, I have a feeling we don't need it as a flag here?

tihuan (approved):
LGTM! Thanks so much for setting this up, @bchu1 🤩 👏 🥇 🙏 !

frontend/packages/data-portal/package.json
bchu1: Forgot to delete this!

bchu1:
> LGTM! Thanks so much for setting this up, @bchu1 🤩 👏 🥇 🙏 !

Thank you for your help! 🙌 😃 🥳

frontend/packages/data-portal/e2e/carryOver.test.ts
```diff
@@ -48,8 +46,6 @@ test.describe('Carry over filters', () => {
       label: translations.objectName,
       value: TEST_VALUE,
     })
-
-    await tableActor.expectResultWithUrlParam(TEST_PARAM, TEST_VALUE)
```
ehoops-cz, line 52: So this is replaced by the snapshot at the end of the test?

ehoops-cz (approved):

frontend/packages/data-portal/e2e/carryOver.test.ts
bchu1: Yep!

bchu1: Yep!

PR #1618, opened 2025-02-12
fix: Disable firefox and webkit e2e tests during prod deploy

bchu1:
Also improve Slack message to only send if the prod deploy action actually ran.


rainandbare (approved):
Thank you Bryan.

PR #1626, opened 2025-02-13
fix: Auto accept Chromatic baseline changes to main

### LINKED ISSUE #1608: Set up GitHub Action

bchu1:
closes #1608 

I think this is necessary because of https://www.chromatic.com/docs/branching-and-baselines/#how-is-merge-base-calculated:~:text=The%20baseline%20is%20the%20last%20accepted%20snapshot%20on%20a%20given%20branch?

Seems like accepting new baselines on PRs only associates the new baselines with the PR's branch? So the action has to run again on `main` for the new baseline to be updated for future runs?

### DIFF
--- a/.github/workflows/frontend-e2e-tests.yml
+++ b/.github/workflows/frontend-e2e-tests.yml
@@ -221,6 +221,7 @@ jobs:
       - name: Run Chromatic
         uses: chromaui/action@latest
         with:
+          autoAcceptChanges: "main"
           playwright: true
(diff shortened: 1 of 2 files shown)


rainandbare:
Ah! Smart - we were struggling with that on cellXgene.

rainandbare (approved):

tihuan:
Great fix! I think using `push` should fix most of the issues where Chromatic couldn't figure out which build to use as new baseline after merging a PR. The `autoAcceptMain` is great to automatically accept any flaky diff that happens on `main` through staging/prod e2e tests I believe!

PR #1631, opened 2025-02-14
Surface neuroglancer in View Tomogram tooltip

### LINKED ISSUE #1283: Fix bug on Runs table

### LINKED ISSUE #1383: Surface Neuroglancer Documentation in the Portal UI and docsite
We'd like to make the Neuroglancer Quickstart guide more discoverable both in the Portal UI and on the doc site to address the UXR finding that most of our users are unfamiliar with using Neuroglancer. 

The Neuroglancer Quickstart documentation is not currently linked to in the Portal UI. @Janeece is there an SDS pattern you'd recommend for this? Linking on or near the View Tomogram buttons would be ideal.

The Neuroglancer Quickstart documentation is located under About the CryoET Data Portal section on the docsite. Renaming that section or moving the Neuroglancer Quickstart to the Tutorials Section are 2 suggestions.

melissawm:
Closes #1383
Closes #1283 

Also addresses an issue where the tooltip was not clickable in the Runs table.

Before:

https://github.com/user-attachments/assets/37f764ee-132c-430a-aec1-043dfdef7841

After this PR:

https://github.com/user-attachments/assets/01ecb249-dfb2-4969-9dab-92304ae407cd


melissawm:
Sorry @rainandbare not sure if you are the right person to review this, feel free to tag others if that's the case. Thanks!

frontend/packages/data-portal/app/components/ViewTomogramButton.tsx
```diff
@@ -38,16 +39,32 @@ export function ViewTomogramButton({
...
     <Tooltip
       tooltip={
-        enabled
-          ? t('viewTomogramInNeuroglancer', {
-              id: `${IdPrefix.Tomogram}-${tomogramId}`,
-            })
-          : t('noTomogramsAvailable')
+        enabled ? (
+          <>
+            <span className="font-semibold">
```
rainandbare, line 44: this can be a div instead of a span and then we can remove the `<br>` tag!

frontend/packages/data-portal/app/components/Dataset/RunsTable.tsx
```diff
@@ -48,8 +48,7 @@ export function RunsTable() {
   const { t } = useI18n()
   const [searchParams] = useSearchParams()
 
-  const [isHoveringOverInteractable, setIsHoveringOverInteractable] =
-    useState(false)
+  const [isHoveringOverInteractable] = useState(false)
```
rainandbare, line 51: If we are not setting this state anywhere I would rather not define it at all! But there seems to be quite a bit of logic based on it. Do you want to explore removing this piece of state entirely?

rainandbare (changes requested):
I can take a closer look at the isHoveringOverInteractable state tomorrow if you don't get to it first!

rainandbare (approved):

frontend/packages/data-portal/app/components/ViewTomogramButton.tsx
melissawm: Looks like you got to it first, thanks!

frontend/packages/data-portal/app/components/Dataset/RunsTable.tsx
melissawm: Thanks for tackling this one - anything else I should do here?

melissawm:
I think this is ready for another look 🙏🏻

kev-zunshiwang:
Thanks Melissa!
Another small nitpik is the hover state of the minimal button should come with a gray background behind it like the screenshot below, you can also refer to this design [here](https://www.figma.com/design/b8QCMLGEEGtd4xuKN14IsB/2025-Q1---UXR-Recommendation?node-id=217-9419&t=p6CCpkfqRq8UTjWJ-1).
My understanding is the sds button should have these all packaged together so its interesting how it's missing the background. Idk how it's coded but I hope we are using the sds minimal button, is this correct?

frontend/packages/data-portal/app/components/Dataset/RunsTable.tsx
```diff
@@ -219,8 +219,8 @@ export function RunsTable() {
               <ViewTomogramButton
                 tomogramId={tomogram?.id.toString()}
                 buttonProps={{
-                  sdsType: 'secondary',
-                  sdsStyle: 'square',
+                  sdsType: 'primary',
+                  sdsStyle: 'minimal',
```
melissawm, line 223: @kev-zunshiwang see here for the sds settings of the minimal button. I will investigate the hover color. Thanks!

kev-zunshiwang:
Screenshot 2025-02-21 at 1 51 44 PM

This is not hovering right, there should be no border and the fill color should be the border color

melissawm:
Ok - changed the PR a bit to meet the design goals, but should be good now. 

My last question is I didn't see specs for the Tomograms table button, should it be left as is? e.g. https://surface-neuroglancer.cryoet.dev.si.czi.technology/runs/16849?table-tab=Tomograms

Screenshot 2025-02-21 at 4 45 04 PM

kev-zunshiwang:
no need to change the tomogram table button!

kev-zunshiwang:
And thanks so much Melissa, this looks great!! No more feedback :)))

rainandbare (approved):
Looks great Melissa!!!! 🔥🔥🔥

melissawm:
Thanks @kev-zunshiwang and @rainandbare for the reviews ❤️

Issue #1641, opened 2025-02-20
Bug: Inconsistent numbers of annotations displayed on Single Deposition Page


uermel:
**Describe the bug**
On the single deposition page, the number of contained annotations is listed in the description section as well as in the table. These numbers are not consistent even though that should be the case.

**To Reproduce**
1. Go to a [deposition page](https://cryoetdataportal.czscience.com/depositions/10318)
2. Compare the number of annotations under `Deposition Data -> Annotations:` and in table column `Annotations -> Deposition only`
 
**Screenshots**
![Image](https://github.com/user-attachments/assets/3f6952c3-3f9b-45d0-b4df-09f217777ca4)

**Desktop (please complete the following information):**
 - OS: MacOS 15.3.1 (24D70)
 - Browser: Chrome 133.0.6943.127

PR #1679, opened 2025-03-05
fix: deposition annotations count

codemonkey800:
#1641

Fixes annotation counts for the dataset table by getting the count from `annotationsAggregate` directly

## Before

image

## After

image


rainandbare (approved):
Looks good Jeremy!

Issue #1688, opened 2025-03-07
Bug: Clicking "View Tomogram" in runs table causes navigation to runs page


uermel:
**Describe the bug**
Clicking "View Tomogram" inside the runs table (single dataset page) opens a new tab with neuroglancer, but also triggers navigation to the single run page in the old tab. 

The expected behavior is that clicking "View Tomogram" only opens the neuroglancer tab, but let's the user remain on the single dataset page in the other tab.

**To Reproduce**
1. Go to any [Single Dataset Page](https://cryoetdataportal.czscience.com/datasets/10446)
2. Click on "View Tomogram" for any run in the table
3. Observe that this causes navigation to the Single Run page as well.

**Screenshots**
https://github.com/user-attachments/assets/7d298133-5f80-4964-8a39-d1a6b8df8baa

**Desktop (please complete the following information):**
 - OS: 15.3.1 (24D70)
 - Browser Chrome 134.0.6998.45 (Official Build) (arm64)

**Additional context**
Users at CZII report that this also causes intermittent crashes with 404 message when browsing many tomograms.

melissawm:
It looks like when the View Tomogram button is clicked, we also capture that click in the corresponding Runs table row. I can take a look.

uermel:
Thank you! fyi, the "View Tomogram" button used to be an actual button with a blue outline. Maybe this broke when it was changed to the current non-outlined version.

rainandbare:
I think I have a fix for this coming @melissawm! Ill tag you in the PR!

melissawm:
Thank you @rainandbare !!

rainandbare:
Happening on the GT tag tooltip too after SDS upgrade

PR #1724, opened 2025-03-21
fix: stop propagation for tooltip links

rainandbare:
Github Ticket: #1688 
Fix to stop bug where you click on a tooltip link and it navigates through to that item in the table. 

Some nits as well!


frontend/packages/data-portal/app/components/ViewTomogramButton.tsx
```diff
@@ -49,6 +49,7 @@ export function ViewTomogramButton({
             <Link
               to={t('neuroglancerTutorialLink')}
               variant="dashed-underlined"
+              stopPropagation
```
rainandbare, line 52: I added this prop to the Link component in an earlier PR so now we can use it here to keep the click from being heard by the parent component and taking us to a new page @melissawm

frontend/packages/data-portal/app/components/AnnotatedObjectsList.tsx
```diff
@@ -53,23 +55,28 @@ export function AnnotatedObjectsList({
...
+                    {t('groundTruthAnnotationAvailable')}{' '}
+                    <Link
+                      className="cursor-pointer text-light-sds-color-primitive-blue-500 border-solid hover:border-b border-light-sds-color-primitive-blue-500"
+                      to={SITE_LINKS.GROUND_TRUTH_FLAG_DOCS}
+                      stopPropagation
+                    >
+                      {t('learnMore')}
+                    </Link>
+                  </>
+                }
```
rainandbare, line 77: Is this an okay I18n compromise @codemonkey800?

PR #1746, opened 2025-04-02
feat: API styling: make attributes list into table

melissawm:
Also includes a fix for the search area that has been missing from the navbar after a recent update of the sphinx-immaterial theme.

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -245,7 +245,11 @@ div.attrstable {
   display: table !important;
   width: 100%;
   margin-bottom: .5em;
-  padding: 0 .8rem;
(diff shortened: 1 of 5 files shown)


dgmccart (approved):
LGTM! Thank you for addressing Janeece's comments. Not sure if this includes the contrast change suggested by Kandarp, but if not, we can address that in a separate PR if needed. Please merge and backport these style changes.

melissawm:
This does not include the dark mode feedback - I think we would require a new design review since those colors/style is the one suggested by Kevin in our latest design review. I will open an issue for it.

dgmccart:
Ok, that sounds good - please add the issue to the design board and ping Kevin on Slack to let him know.

PR #1767, opened 2025-04-09
fix: Fix PR preview deployments to docs site

melissawm:
I just realized that instead of just running the build docs job for pull requests, I was also deploying the site to the dev folder 😬 

This adds a guard to prevent the deployment if we are in a pull request.


rainandbare (approved):
😬

Issue #1740, opened 2025-03-27
Update how is frame available is computed in frontend


manasaV3:
### Description

The frames available info in the run page, is determined by checking if the count of frames is greater than 0 (ref : [here](https://github.com/chanzuckerberg/cryoet-data-portal/blob/e11ee135fb8c0cb0ece137489e43b8fe621d0dec/frontend/packages/data-portal/app/components/Run/RunHeader.tsx#L117)). 

This worked in the past, as  we would generate a frame record only when there are frame files. But, we now support the addition of frame records even when there are no files. So, having frame records doesn't necessarily mean there are frame files. 

We need to update the related graphql query that fetches frames data, to include an additional filter, to only fetch frame records where `httpsFramePath` or `s3FramePath` is not null.

PR #1786, opened 2025-04-23
fix: frames available

codemonkey800:
#1740

Updates query for getting frame data by fetching list of frames with non-null file paths instead of using the frame aggregate count.


rainandbare (approved):
Thank you for this!

Issue #1774, opened 2025-04-16
Tilt series quality filter does not work correctly when selecting multiple quality levels.


uermel:
**Describe the bug**
When selecting multiple quality scores in the tilt series quality filter on the single dataset page, only those runs of the first score that was selected are displayed. The filter is expected to act as an OR filter. 

**To Reproduce**
1. Go to [dataset 10045](https://frontend.cryoet.staging.si.czi.technology/datasets/10045)
2. Select `3 - Moderate`, 5 runs are displayed
3. Select `4 - Good` in addition, the same 5 runs remain displayed (59 expected)

Works in reverse as well: 
1. Go to [dataset 10045](https://frontend.cryoet.staging.si.czi.technology/datasets/10045)
2. Select `4 - Good`, 54 runs are displayed
3. Select `3 - Moderate` in addition, the same 54 runs remain displayed (59 expected)

**Screenshots**
https://github.com/user-attachments/assets/75319c08-7dc7-414d-af15-05cc017cadf6

**Desktop (please complete the following information):**
OS: MacOS 15.4 (24E248)
Browser Chrome Version 135.0.7049.96 (Official Build) (arm64)

PR #1787, opened 2025-04-23
fix: tilt series quality score filter for multiple values

codemonkey800:
#1774

Fixes issue with the quality score filter when selecting multiple values. The issue is that passing `parseInt()` directly into the `map()` function behaves erroneously because it uses the 2nd argument of `map()` as the radix input for `parseInt()`:

image

image

## Before

image

## After

image


rainandbare:
> The issue is that passing parseInt() directly into the map() function to behave erroneously because it uses the 2nd argument of map() as the radix input for parseInt()

lol wild! Thank you for this fix!

rainandbare (approved):

bchu1 (approved):

PR #1791, opened 2025-04-24
fix: view tomogram button propogration

codemonkey800:
Follow up of #1782 to fix the `View tomogram` plausible event not firing on click. Instead of stopping propogation on the link itself, we can stop it in the outer div used for logging the plausible event. This will also allow us to restore the view tomogram button as a link which is semantically preferred.


rainandbare (approved):
Reaallllll nice! Thanks for catching those important details!

PR #1798, opened 2025-05-05
fix: Fix version switcher behavior

### LINKED ISSUE #1754: docs: Verify that version switcher is reaching the right pages
**Describe the bug**
The version switcher seems to be redirecting to the main page of the docs site instead of the current page. 

**To Reproduce**
1. Go to any page in the docs site that is not the landing page;
2. Click on the version switcher and select a different version from the current one;
3. Page redirects to the landing page, instead of a different version of the current page.

melissawm:
Closes #1754

PREVIEW: https://melissawm.github.io/cryoet-data-portal/dev/data_model.html#data-model

To test:

- Verify that clicking on the version dropdown only opens the dropdown and doesn't navigate to the main portal page;
- Verify that for any pages/subpages, clicking on a different version in the dropdown correctly redirects to the same url, but with the selected version.
Example: navigating to
https://melissawm.github.io/cryoet-data-portal/dev/data_model.html#data-model
And clicking on the "Stable" version in the switcher should bring you to 
https://melissawm.github.io/cryoet-data-portal/v4.0/data_model.html#data-model

### DIFF
--- a/docs/_static/js/version_redirect.js
+++ b/docs/_static/js/version_redirect.js
@@ -1,0 +1,11 @@
+window.addEventListener("DOMContentLoaded", function () {
+  versions = document.querySelectorAll(".md-version__item");
+  const currentUrl = window.location.href;
+  var currentVersion = currentUrl.match(/\/v\d+\.\d+(\-?rc\d+)?\/|\/dev\//g)[0];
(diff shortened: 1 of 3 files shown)


melissawm:
cc @dgmccart

dgmccart (approved):

Issue #1799, opened 2025-05-05
docs: Make "latest" the default version for the docs site


melissawm:
As discussed in slack.

melissawm:
This has been done together with #1798

PR #1800, opened 2025-05-05
docs: Add documentation on deployment and versioning of docs site

### LINKED ISSUE #1777: docs: Add README with versioning and branching instructions for docs site
To make sure the version switcher works properly and that versions of documentation match the API, we will need to keep versioned branches of the docs site. 

We will need to add instructions on how to do this. The simplest way to make this visible is to have a README in the docs folder of the repo.

melissawm:
Closes #1777
Also addresses #1799


melissawm:
Once this PR is merged, I'll update the symlink on gh-pages to point to latest.

dgmccart (approved):

Issue #1742, opened 2025-03-28
The "Author ORCID" filter for Depositions is not working correctly


cnguyenvse:
**Describe the bug**
The Author ORCID filter for Depositions returns the same total number of depositions no matter what ORCID is filtered with.

1. Go to 'Deposition' [page](https://cryoetdataportal.czscience.com/browse-data/depositions) 
2. Click on 'Deposition Author' in the left 'Filter by' section
3. Scroll down to 'Author ORCID' and enter any ORCID, 
 example (Author Name: Elizabeth Montabana      Author ORCID:  0000-0002-6731-5854)
4. Notice the Deposition table updates but the results are not filtered

**Screenshots**

<img width="1247" alt="Image" src="https://github.com/user-attachments/assets/a385a30c-f2a0-4878-834d-9b903a531b32" />

**Desktop (please complete the following information):**
 - OS: [MacOS]
 - Browser [Chrome]

cnguyenvse:
@junxini  Verified fix in prod

<img width="1087" alt="Image" src="https://github.com/user-attachments/assets/e67b6d5c-f0af-4d80-811e-03257d0a69db" />

PR #1817, opened 2025-05-20
fix: depositions orcid filterby

rainandbare:
Ticket: [#1742](https://github.com/chanzuckerberg/cryoet-data-portal/issues/1742)

This fix means that you can now

- [x] find a deposition using an orcid 
- [x] find a deposition using an orcid & kaggleId (in the name field)
- [x] find a deposition using an orcid & name

Screenshot 2025-05-20 at 10 25 58 AM


rainandbare:
> oh man wow so we never added it 🤣 glad it was a 1-liner, thank you for this!

🤦‍♀️ The merging of the kaggleId and the name fields was a relatively last minute change, before we added that the orcid would have worked. But obviously did not afterwards!

Issue #1292, opened 2024-10-31
Results make no sense when filtering by Deposition ID displayed in Tomogram Details applet


lauramarcos:
**Describe the bug**
Results make no sense when filtering by Deposition ID displayed in Tomogram Details applet

**To Reproduce**
1. Go to https://cryoetdataportal.czscience.com/runs/364?table-tab=Tomograms&metadata=tomogram&tab=metadata
2. Copy Dataset ID and go to Browse Data
3. In Deposition ID filter, enter that Deposition ID previously copied
4. See there are no results

**Screenshots**

https://github.com/user-attachments/assets/d83f049b-0ebe-486d-a7dc-8eeb29532418

**Desktop (please complete the following information):**
 - OS: Sonoma 14.7
 - Browser Chrome

junxini:
here's another example: https://cryoetdataportal.czscience.com/runs/16470?deposition-id=10310&table-tab=Tomograms&metadata=tomogram&tab=metadata

you'll notice that the deposition ID field in the info panel is not accurate. It should say CZCDP-10310

PR #1835, opened 2025-06-04
fix: deposition id prefix

codemonkey800:
#1292

Fixes the prefix we use for the deposition ID

### Before

image

### After

image


rainandbare (approved):
Nice! A one liner! For so much confusion I am glad it was simple enough.

Issue #1825, opened 2025-05-23
Download Tomogram dialog selection focus not working correctly


cnguyenvse:
**Describe the bug**
On the Download Tomogram dialog, when the user selects "Download Tomogram"  or "Download all Annotations"  the background turns grey as expected but the corresponding radio button on the left remains unselected, it should turn blue.  

**To Reproduce**
1. Go to any single Run page
2. Click on '(Download...)' in the upper right hand corner
3. The 'Configure Download' dialog box should appear
4. Select "Download Tomogram"  or "Download all Annotations" options
5. Notice the option's background turns grey but the radio button remains unselected
6.  Click on the radio button of the unselected option

**Screenshots**
Here none of the radio buttons are selected even though "Download Tomogram" is selected and the background is grey 
<img width="1312" alt="Image" src="https://github.com/user-attachments/assets/f5775b5c-e0ff-473a-af47-27e495ab01ae" />

Here "Download Tomogram" is selected and the background is grey but the "Download all Annotations" radio button is selected in blue
<img width="1312" alt="Image" src="https://github.com/user-attachments/assets/b8f18ba8-6724-43e4-9042-c160682b62f9" />

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome

PR #1836, opened 2025-06-04
ci: fix E2E tests

codemonkey800:
- The E2E tests were failing due to #1825
  - Because the radio was not checking, the tests were failing
  - Still unsure of the root cause, the issue only happens in staging / prod
  - Unable to reproduce in localhost using both docker and prod builds
- Adds temporary fix for above issue by checking the radio when selecting the download configuration

## Demo

This run contains a passing 

https://github.com/chanzuckerberg/cryoet-data-portal/actions/runs/15449894037

### DIFF
--- a/frontend/packages/data-portal/e2e/downloadDialog.test.ts
+++ b/frontend/packages/data-portal/e2e/downloadDialog.test.ts
@@ -244,7 +244,6 @@ test.describe('downloadDialog', () => {
 
         await downloadDialogPage.clickDialogRadio(
           translations.downloadAllAnnotations,
-          DownloadConfig.AllAnnotations,
         )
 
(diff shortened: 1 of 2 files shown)


rainandbare (changes requested):
I was able to recreate on local and the issue is happening on this line:

https://github.com/chanzuckerberg/cryoet-data-portal/blob/e6297f47befc69e14689e9f9cd86c466777a1141/frontend/packages/data-portal/app/components/Download/ConfigureTomogramDownloadContent.tsx#L47

Could you add some logic to set the downloadConfig when we open the Download dialogue? Right now downloadConfig is undefined so none of the Radio Buttons are selected. 

Thank you!! Then we will have dealt with the root cause.

rainandbare (approved):

PR #1845, opened 2025-06-09
chore: Update example IDs to more realistic number

### LINKED ISSUE #1381: Further improve python API client docs for get_by_id
The python API client now has `Model`-specific docs for every `Model` it supports. While the examples for `get_by_id` do contain working code, it would be great if the queries they imply would actually return data instead of `None`. This is done in the case of `find` to some extent.

Right now, for the majority of `Models`, the example in `Model.get_by_id` uses `id=1`, which will return None for most models, because that ID in most cases is not used. 

To reproduce: 
```python
import sys
import inspect

classes = [obj for name, obj in inspect.getmembers(sys.modules[cdp._models.__name__], inspect.isclass)
          if obj.__module__ is cdp._models.__name__]

client = cdp.Client()

for c in classes:
    ret = c.get_by_id(client, 1) is None
    print(f"{c.__name__}.get_by_id(client, 1) is None == {ret}")
```

Returns: 
```
Alignment.get_by_id(client, 1) is None == True
Annotation.get_by_id(client, 1) is None == True
AnnotationAuthor.get_by_id(client, 1) is None == True
AnnotationFile.get_by_id(client, 1) is None == True
AnnotationMethodLink.get_by_id(client, 1) is None == True
AnnotationShape.get_by_id(client, 1) is None == True
Dataset.get_by_id(client, 1) is None == True
DatasetAuthor.get_by_id(client, 1) is None == True
DatasetFunding.get_by_id(client, 1) is None == True
Deposition.get_by_id(client, 1) is None == True
DepositionAuthor.get_by_id(client, 1) is None == False
DepositionType.get_by_id(client, 1) is None == True
Frame.get_by_id(client, 1) is None == True
FrameAcquisitionFile.get_by_id(client, 1) is None == True
GainFile.get_by_id(client, 1) is None == True
PerSectionAlignmentParameters.get_by_id(client, 1) is None == False
Run.get_by_id(client, 1) is None == True
TiltSeries.get_by_id(client, 1) is None == True
Tomogram.get_by_id(client, 1) is None == True
TomogramAuthor.get_by_id(client, 1) is None == True
TomogramVoxelSpacing.get_by_id(client, 1) is None == True
```

melissawm:
Closes #1381 

I ran `make codegen` and `make build` locally, not sure if I should edit just the jinja file though.

### DIFF
--- a/client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
+++ b/client/python/cryoet_data_portal/src/cryoet_data_portal/_models.py
@@ -116,7 +116,7 @@ class Alignment(Model):
         Examples:
             Get an Alignment by ID:
 
-            >>> alignment = Alignment.get_by_id(client, 1)
+            >>> alignment = Alignment.get_by_id(client, 10000)
             >>> print(alignment.id)
         """
         return super(Alignment, cls).get_by_id(client, id)
@@ -226,7 +226,7 @@ class Annotation(Model):
         Examples:
             Get an Annotation by ID:
 
-            >>> annotation = Annotation.get_by_id(client, 1)
+            >>> annotation = Annotation.get_by_id(client, 10000)
             >>> print(annotation.id)
         """
         return super(Annotation, cls).get_by_id(client, id)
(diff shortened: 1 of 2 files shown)


melissawm:
Thanks @kaloster - I'll merge, but if anything is not right I'll come back to fix :) Cheers!

PR #1856, opened 2025-06-19
feat: Add a citation button to pages

melissawm:
There may be some cleanup to be done later, as I've based this implementation on the View Tomogram button.

The implementation of plausible tracking is still pending.


frontend/packages/data-portal/app/components/Dataset/DatasetHeader.tsx
```diff
@@ -25,7 +26,7 @@ export function DatasetHeader() {
   return (
     <PageHeader
       actions={
-        <div>
+        <div className="flex items-center gap-2.5">
```
rainandbare, line 29: let's use `gap-sds-s` to keep our spacing scale explicitly in line with sds recommendations!

rainandbare (approved):

melissawm:
@dgmccart Please take a look when you have the time. Thanks!

frontend/packages/data-portal/app/components/Dataset/DatasetHeader.tsx
melissawm: @rainandbare hopefully this is what you meant!

Issue #1755, opened 2025-04-08
docs: Update data organization page


dgmccart:
Pending new graphics from brand and creative team + copy changes from Dannielle
- [ ] Add updated data schema overview
- [ ] Add updated run download file structure
- [ ] Mention frames/ctf article in relevant section (tomograms ?)
- [ ] Expand tomogram section to explain tags
- [ ] Add information about GT tag in annotations table

dgmccart:
I can work on the copy changes for a separate PR.

dgmccart:
The copy changes mentioned in the checklist above haven't been added, but that is waiting on me. I moved this to the waiting status.

PR #1857, opened 2025-06-20
docs: Update data schema and run download figures

melissawm:
Partly addresses #1755


dgmccart:
Could you add a preview so that I can check how it renders?

dgmccart:
Screenshot 2025-06-26 at 1 45 19 PM
The run download appears to be the old image.

melissawm:
Wow - sorry about that, I think I must have copied the image from the wrong folder!

I've just updated the preview, please check again.

PR #1865, opened 2025-06-23
feat: Add full changelog of Python Client to docs site

melissawm:
This PR adds the full changelog for the Python API client to the docs site. This is done by copying over the existing changelog file for the client into the docs sources and adding it as a new markdown page.

It also renames the current "Changelog" page to "Release notes" which is more appropriate since it contains highlights of major releases only.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,6 +14,12 @@ help:
 
 .PHONY: help Makefile
 
+changelog:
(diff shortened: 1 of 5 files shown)


melissawm:
This is how the Python API main page looks with this PR:

Screenshot 2025-06-23 at 12 45 29 PM

And this is what the full changelog page looks like:

Screenshot 2025-06-23 at 12 46 06 PM

melissawm:
Thanks, both!

PR #1911, opened 2025-07-28
feat: Adding cross modality fields to schema

manasaV3:


.github/workflows/client-python.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
```
uermel, line 24: Why are we removing ubuntu from the tests? The client should absolutely support linux systems.

uermel (changes requested):
Please still test the client on ubuntu. Otherwise looks fine.

.github/workflows/client-python.yml
manasaV3: Totally agree, was trying to debug some failing tests.

client/python/cryoet_data_portal/tests/test_filters.py
```diff
@@ -20,19 +20,15 @@ def test_basic_filters(client) -> None:
 def test_filter_merge(client) -> None:
     # Make sure our GQL filters get merged instead of letting the longest-path
     # queries overwrite shorter paths.
-    runs = Run.find(client, [Run.name == "RUN1"])
-    assert len(runs) == 1
-    assert runs[0].name == "RUN1"
```
uermel, line 25: Are these parts of the test not necessary?

manasaV3: No, it is not, as it is not testing the filters being merged. It is covered in `test_basic_filters` just above this test.

PR #1917, opened 2025-07-28
feat: GraphQL and API infrastructure for depositions group-by

codemonkey800:
## Summary
- Implements complete data fetching layer for depositions grouping functionality
- Adds 12 new API routes for efficient data aggregation
- Creates GraphQL server functions for dataset and run data fetching
- Builds comprehensive utility functions for data transformation
- Adds core data fetching hook with extensive testing coverage

🤖 Generated with [Claude Code](https://claude.ai/code)


codemonkey800:
ignore the E2Es, they should be passing in the last PR #1919

frontend/packages/data-portal/app/queries/useDepositionRunsForDataset.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param params - Object containing depositionId, datasetId, page, type, and optional enabled flag
+ * @returns React Query result with runs data, loading state, and error handling
+ */
+export function useDepositionRunsForDataset({
+  depositionId,
+  datasetId,
+  page,
+  type,
+  enabled = true,
```
vincent-czi, line 20: Okay, so I don't see this hook being used anywhere in the code for this PR, so I'm assuming its usage comes in a later PR. But, I'm a bit confused by this `enabled` flag: what's its purpose? It looks like it plays a small role in the TS types (maybe? I'm not that great with TS types and I haven't dug in deeply here.), but for its real impact on logic it just gets passed right through with nothing done to it or affected by it? Am I correct on that?

It might be that it makes more sense in the context of how `useDepositionRunsForDataset` gets used later on, but it seems odd to have a param passed in that doesn't do anything other than come out the other end as part of the returned object.

I see similar `enabled = true` function params in this PR. Let this comment be a stand in for asking about all of them. This is more in the class of a nit: I don't think I really have enough context on what the ultimate use-case will be to really understand what's going on. But if it seems weird to you in retrospect as well, maybe figure out a way to eliminate it entirely?

frontend/packages/data-portal/app/queries/useDatasetsForDeposition.ts
```diff
@@ -0,0 +1,92 @@
...
+  DataContentsType,
+  DepositionDatasetsResponse,
+} from 'app/types/deposition-queries'
+import { isDefined } from 'app/utils/nullish'
+
+/**
+ * Hook to fetch datasets for a deposition using React Query + API route.
+ *
+ * This uses a server-side API route to avoid CORS issues with direct GraphQL calls,
+ * leveraging React Query for caching and state management.
```
vincent-czi, line 17: This doesn't seem crazy to do, but we could also tweak the BE server to put in the appropriate CORS headers. Shouldn't be too hard to do. But I can also see this approach being reasonable in-and-of-itself, not simply a workaround.

If we ever want to make the BE be directly callable by the FE and deal with the CORS headers directly, reach out in the future and we can work on it.

frontend/packages/data-portal/app/routes/api.deposition-datasets.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (!depositionId || Number.isNaN(Number(depositionId))) {
+    return new Response('Missing or invalid depositionId', { status: 400 })
+  }
+
+  if (
+    !type ||
+    !Object.values(DataContentsType).includes(type as DataContentsType)
+  ) {
+    return new Response(
+      'Missing or invalid type parameter. Must be "annotations" or "tomograms"',
```
vincent-czi, line 37: Nit: Would be nice if it wasn't a literal string with `"annotations" or "tomograms"` in it, but rather was interpolated from the values of `DataContentsType`. Dunno if that's a huge pain to do though. I'm just thinking that this error message could get stale if we eventually support a third option, but this error message doesn't get updated.

frontend/packages/data-portal/app/routes/api.deposition-datasets.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return new Response(JSON.stringify(response), {
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error('Failed to fetch datasets for deposition:', error)
```
vincent-czi, line 85: Pure question for curiosity: is `console.error` how we get error logging to show up in CloudWatch (or somesuch) for when we have server side errors?

frontend/packages/data-portal/app/utils/api-helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Handles API errors with consistent formatting and logging
+ * @param error The error that occurred
+ * @param context A description of where the error occurred
+ * @returns A standardized error response
+ */
+export function handleApiError(error: unknown, context: string): Response {
+  // Use structured logging if available, otherwise fall back to console.error
+  // eslint-disable-next-line no-console
+  console.error(`${context}:`, error)
```
vincent-czi, line 47: I feel like this `handleApiError` function is very similar to a lot of code blocks in the loaders I see elsewhere that go roughly like this:
``` try {
  } catch (error) {
    // eslint-disable-next-line no-console
    console.error('Failed to fetch tomograms for run and deposition:', error)
    return new Response('Failed to fetch tomograms for run and deposition', {
      status: 500,
    })
```

That code block above repeats often in the loaders, could we swap over to using this function? That said, this function is a bit different since the `Failed to` preamble appears in both in the above, while it only appears in the Response for the function. But I think it would make sense to standardize and have them all use the same `handleApiError`.

vincent-czi (approved):
Looks good! A lot to review in one go, but I read everything and nothing serious jumped out at me. Most of the comments are nits, so I'm fine to approve. Wouldn't be an issue if it got merged now, although see if any of the comments strike you as valuable first.

frontend/packages/data-portal/app/routes/api.deposition-datasets.ts
codemonkey800: yeah all console logs whether it be `console.warn()` or `console.error()` get tracked in CloudWatch 🫡

PR #1920, opened 2025-07-29
fix: Updating the workflow dispatch to deploy docsite

manasaV3:
## Description

Currently, we require an update to the docs folder for the docsite to be deployed. This makes it cumbersome to update the doc-site for new client versions. 

The changes allow the doc-site to be deployed with manual trigger of the worklfow.

### DIFF
--- a/.github/workflows/docsite-build-deploy.yml
+++ b/.github/workflows/docsite-build-deploy.yml
@@ -57,12 +57,12 @@ jobs:
         run: |
           cd docs/
           make clean && make html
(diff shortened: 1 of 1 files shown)


melissawm (approved):
I think we can already do this (see line 3 of the workflow file and the screenshot)

Screenshot 2025-07-29 at 9 03 36 AM

This is actually how I do backports to the docs. However I'm now unsure so I think this is safer anyways 😄

PR #1969, opened 2025-09-15
chore: temporarily downgrade dependencies to fix E2E tests

codemonkey800:
## Summary
Temporarily removes security overrides for form-data and path-to-regexp dependencies that were recently added to resolve dependabot alerts, but are causing E2E test failures.

## Changes
- Remove `form-data: ^4.0.4` security override from frontend/package.json
- Remove `path-to-regexp: ^0.1.12` security override from frontend/package.json
- Update pnpm-lock.yaml to reflect dependency changes

## Notes
This is a temporary solution to unblock PRs. The long-term plan is to:
1. Restore the patched versions
2. Fix the E2E tests to work with the updated dependencies
3. Re-apply the security patches once tests are compatible

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -24,8 +24,6 @@
       "braces": "^3.0.3",
       "cross-spawn": "^7.0.5",
       "dset": "^3.1.4",
-      "form-data": "^4.0.4",
-      "path-to-regexp": "^0.1.12",
       "rollup": "^3.29.5",
       "vite": "^4.5.3",
       "ws": "^7.5.10"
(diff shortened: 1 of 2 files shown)


codemonkey800:
> Uh oh - are the dependancy checks going to fail now because of vulnerable packages?

@rainandbare yeah unfortunately these are going to fail, but only for this PR since it's the one introducing the dependencies. they usually only run when a package is added or upgraded, so it shouldn't affect other PRs

codemonkey800:
need to figure out a proper fix for this, but right now this is blocking a few PRs from being merged so reverting it for now will unblock those PRs

PR #2050, opened 2026-03-19
fix: pass environment-specific API_URL_V2 during Docker build and CI codegen

PALLAVIKHEDLE:
### Problem
During Docker image builds and CI test runs, the pnpm build:codegen step (which generates GraphQL types) was not being passed the correct GraphQL API endpoint. This caused codegen to either fail or generate types against the wrong environment's schema.

### Changes

- `dockerfile`:  Accept API_URL_V2 as a build arg and expose it as an env variable so pnpm build:codegen uses the correct endpoint during the Docker build step.
- `docker-build-rdev.yaml`: Passes the staging GraphQL URL  for rdev builds.
- `docker-build-staging-prod.yaml`:  Split the single argus_builder job into two separate jobs (build_staging, build_prod) so each env gets its own correct API URL.

### DIFF
--- a/.github/workflows/docker-build-rdev.yaml
+++ b/.github/workflows/docker-build-rdev.yaml
@@ -19,3 +19,4 @@ jobs:
       envs: rdev
       path_filters: 'frontend/**'
       branches_ignore: main,release-please--branches--main--components--cryoet-data-portal-frontend
+      api_url_v2: 'https://graphql.cryoet.staging.si.czi.technology/graphql'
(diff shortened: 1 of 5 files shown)


manasaV3 (changes requested):
The pattern is to promote an image validated in staging to the prod environment. This breaks that pattern. Is there a reason, we can't set the environment variable for the test to point to the relevant api url?

PALLAVIKHEDLE:
> The pattern is to promote an image validated in staging to the prod environment. This breaks that pattern. Is there a reason, we can't set the environment variable for the test to point to the relevant api url?

Totally agree with the pattern - I had the same concern.
The `frontend-tests.yml` already sets API_URL_V2 from the GitHub environment variables for the codegen step there-so that part is covered.
The tricky bit is that codegen also runs inside the Docker build, so we need the URL available at build time, not just test time. We can't inject it after the image is built. That's the reason  I went with separate builds - but happy to revisit if you have a better idea for handling the build-time dependency!

PR #2096, opened 2026-07-03
fix: detect submitted alignments by per-section params, not method type

daniel-ji:
## Problem
The Dataset Contents summary shows "Alignment: Not Submitted" for datasets whose alignments have `method_type: undefined` (which the db-import maps to a null `alignmentMethod`), even when real per-section alignment data exists. This affects all simulated / undefined-method datasets (e.g. 10439, 10441, and the CryoETSim datasets 10490-10492).

## Fix
Count an alignment as submitted when it has per-section alignment parameters (`perSectionAlignmentsAggregate.count > 0`) rather than when `alignmentMethod` is non-null. Auto-generated default/identity placeholder alignments have zero per-section params, so they still correctly read as "Not Submitted", while real alignments (any method, including `undefined`) now show as "Available".

Verified against the staging API: the filter returns all real alignments and excludes zero-parameter placeholders.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.infra/rdev/values.yaml
+++ b/.infra/rdev/values.yaml
@@ -2,7 +2,7 @@ stack:
   services:
     frontend:
       image:
-        tag: sha-f7878c6
+        tag: sha-3eda32d
       replicaCount: 1
       env:
         - name: API_URL_V2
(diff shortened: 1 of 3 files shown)


daniel-ji:
Closing in favor of a cleaner approach: add a 'simulated' value to the alignment method-type enum (keeps the existing non-null method check; simulated alignments become legitimately non-null) instead of changing the frontend query. Keeping the branch for reference.

